PR #11, opened 2023-02-25
Feature -> Multiple alarms management

gdelataillade:
The purpose of this pull request is to add the management of multiple alarms. This way, you will be able to set and stop multiple alarms at the time, which will be associated with a unique id.

Please note that the development of this feature is challenging so it will take some time. Any help is welcome.


gdelataillade (commented):
OK

gdelataillade (commented):
OK

PR #25, opened 2023-03-30
Fix -> iOS crash when invalid audio url

LesYampolskyi:
The purpose of this pull request is to fix ios app crash when you set invalid audio url.

Description: If you set invalid assetAudio link for <code>setAlarm</code> method it handles in try - catch block and calls <code>result(FlutterError.init(...))</code>. But it doesn't stop execution of <code>setAlarm</code>. And at the next unwraping <code>let currentTime = self.audioPlayers[id]!.deviceCurrentTime</code> app crashes.


gdelataillade:
Hi @LesYampolskyi 

Thank you for your interest in the package.

This looks nice. I'll test your code and merge it as soon as possible.

gdelataillade (approved):
OK

PR #26, opened 2023-03-31
Prevent unnecessary callbacks in ios

LesYampolskyi:
The purpose of this pull request is prevent unnecessary callbacks from DispatchQueue.main.asyncAfter. 

If you stop an alarm or set a new one instead the previous the callback from DispatchQueue.main.asyncAfter will be executed anyway. There is [DispatchWorkItem](https://stackoverflow.com/questions/44633729/stop-a-dispatchqueue-that-is-running-on-the-main-thread) allowing to stop their execution.


gdelataillade:
Great! Thank you I'll merge it soon.

gdelataillade (approved):

PR #39, opened 2023-04-14
Stop the alarm on open notification

LesYampolskyi:
The purpose of this pull request is to stop the alarm's ringing/vibration when the user opens the local notification.

### DIFF
--- a/lib/alarm.dart
+++ b/lib/alarm.dart
@@ -9,7 +9,6 @@ import 'package:alarm/src/android_alarm.dart';
 import 'package:alarm/service/notification.dart';
 import 'package:alarm/service/storage.dart';
 import 'package:flutter/foundation.dart';
-import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 
 /// Custom print function designed for Alarm plugin.
 DebugPrintCallback alarmPrint = debugPrintThrottled;
@@ -39,21 +38,10 @@ class Alarm {
 
     await Future.wait([
       if (android) AndroidAlarm.init(),
-      AlarmNotification.instance.init(
-        onSelectNotification,
-      ),
+      AlarmNotification.instance.init(),
       AlarmStorage.init(),
     ]);
     await checkAlarm();
-  }
-
-  // Stop the alarm when the notification is opened.
-  static onSelectNotification(NotificationResponse notificationResponse) async {
-    if (notificationResponse.id != null) {
-      await Alarm.stop(
-        notificationResponse.id!,
-      );
-    }
   }
 
   /// Checks if some alarms were set on previous session.
--- a/lib/alarm.dart
+++ b/lib/alarm.dart
@@ -9,6 +9,7 @@ import 'package:alarm/src/android_alarm.dart';
 import 'package:alarm/service/notification.dart';
 import 'package:alarm/service/storage.dart';
 import 'package:flutter/foundation.dart';
+import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 
 /// Custom print function designed for Alarm plugin.
 DebugPrintCallback alarmPrint = debugPrintThrottled;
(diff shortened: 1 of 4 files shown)


gdelataillade:
Thank you @LesYampolskyi !

Do you think this has to be the default behavior ? Shouldn't we add a parameter to disable it ? What do you think ?

lib/alarm.dart
```diff
@@ -9,6 +9,7 @@ import 'package:alarm/src/android_alarm.dart';
 import 'package:alarm/service/notification.dart';
 import 'package:alarm/service/storage.dart';
 import 'package:flutter/foundation.dart';
+import 'package:flutter_local_notifications/flutter_local_notifications.dart';
```
gdelataillade, line 12: I would prefer if the `flutter_local_notifications` code is only called from the `AlarmNotification` service. Would that be possible ?

LesYampolskyi: Yes, it is totally possible.

I imported flutter_local_notifications package to lib/alarm only for the NotificationResponse type in onSelectNotification. Any code from flutter_local_notifications is not used.

I thought about circular imports. lib/alarm uses lib/services/notification and that's why i placed onSelectNotification in lib/alarm.
But i see that lib/services/notification also uses lib/alarm (alarmPrint) and [the circular dependencies are allowed for dart](https://stackoverflow.com/questions/65324492/dart-circular-imports)

I replaced the callback and added a new commit.

LesYampolskyi:
> 

Maybe it is better to make it changeable and add a parameter to disable it. By default, it is true, but the developer can change it. For example, if the business logic requires some extra steps for the user after opening the app and doing something manually.
On Android and iOS alarms, there is also a snooze option. Perhaps this one should be added as well.

lib/service/notification.dart
```diff
@@ -17,22 +17,42 @@ class AlarmNotification {
...
+    final initializationSettingsIOS = DarwinInitializationSettings(
       requestAlertPermission: false,
       requestBadgePermission: false,
       requestSoundPermission: false,
+      onDidReceiveLocalNotification: (
+        int? id,
+        String? title,
+        String? body,
+        String? payload,
+      ) async {},
```
gdelataillade, line 32: This callback is empty. Is it possible to remove it so we can make `initializationSettingsIOS` a const again ?

lib/alarm.dart
gdelataillade: Awesome thank you.

lib/service/notification.dart
LesYampolskyi: This callback is only applicable to iOS versions older than 10. 
I've added a <code>onSelectNotificationOldIOS</code> to this callback. It uses a different set of arguments so I can't reuse <code>onSelectNotification</code>

LesYampolskyi:
@gdelataillade how should the parameter be named? Is <code>stopAlarmOnReceivedNotification</code> ok?

And I noticed in <code>alarm_settings.dart</code> that <code>enableNotificationOnKill</code> is missing in <code>copyWith</code>, <code>toJson</code>  and <code>factory AlarmSettings.fromJson</code>. As far as I understand they used to recreate the alarms after app is killed and should this parameter be added as well?

gdelataillade:
Yes `enableNotificationOnKill` is missing, I forgot to add it thanks !

I suggest to call the parameter `stopOnNotificationOpen`. What do you think ?

LesYampolskyi:
> Yes `enableNotificationOnKill` is missing, I forgot to add it thanks !
> 
> I suggest to call the parameter `stopOnNotificationOpen`. What do you think ?

Sounds good.

LesYampolskyi:
@gdelataillade I added the parameter <code>stopOnNotificationOpen</code> for <code>AlarmSettings</code> and modified <code>README.md</code>. And also added the missed <code>enableNotificationOnKill</code>.

I

gdelataillade (approved):

gdelataillade:
@LesYampolskyi sounds good

Have you tested on Android ?

LesYampolskyi:
> @LesYampolskyi looks good
> 
> Have you tested on Android ?

Yes, it works

gdelataillade:
Available in `1.0.3` update.

PR #45, opened 2023-04-27
Enable custom time zone selection for notifications

r0227n:
Allow the TimeZoneId to be freely set, enabling the notification time zone to be customized. Currently, the time zone is fixed to UTC, which makes it inconvenient to use, so I made this modification."

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ assetAudio |   `String`     | The path to you audio asset you want to use as rin
 loopAudio |   `bool`     | If true, audio will repeat indefinitely until alarm is stopped.
 vibrate |   `bool`     | If true, device will vibrate indefinitely until alarm is stopped. If [loopAudio] is set to false, vibrations will stop when audio ends.
 fadeDuration |   `double`     | Duration, in seconds, over which to fade the alarm volume. Set to 0 by default, which means no fade.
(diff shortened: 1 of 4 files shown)


gdelataillade:
Hi @r0227n 

Thanks for your work ! I'm not sure to understand very well the issue though. You had issues with the time the notification is shown ? It is supposed to be scheduled according to the local time of the device.

r0227n:
@gdelataillade 
I'm sorry, it was my mistake. The notifications are working correctly now.

Could you help me with something? Is it intentional that notifications don't appear when the app is running? I would like to have notifications displayed even when the app is active, in addition to the alarm sound.

https://user-images.githubusercontent.com/40818362/235159100-25d67ccd-9fdd-4450-bdb6-79bfeaff2b0b.mp4

gdelataillade:
Hi @r0227n 

Yes it was intentional that the notification does not appear when app is on foreground. I thought it was not especially relevant to show a notification. However, if you convince me, I can change it.

r0227n:
Thank you @gdelataillade 
I understand that it's not possible to show a notification when the app is in the foreground.
Since a specific action can be taken when the time comes, I don't think it's a problem.

Close this PR.

Issue #66, opened 2023-07-18
Alarm sound is playing in delay of 5 seconds


dtkdt100:
**Alarm plugin version**
latest

**Describe the bug**
Alarm sound is playing in delay of 5 seconds. I am setting an alarm for DateTime.now() and only after 3 seconds the alarm rings

**To Reproduce**
*Steps to reproduce the behavior:*
```dart
final alarmSettings = AlarmSettings(
      id: _ringingId,
      dateTime: DateTime.now(),
      assetAudioPath: assetAudioPath,
      loopAudio: false,
      vibrate: LocalMemory.vibrationOnAlarm!.value,
      enableNotificationOnKill: false,
    );
 await Alarm.set(alarmSettings: alarmSettings);
```

**Expected behavior**
The alarm needs to go off in the same time it setts. Maybe 1 second delay

What I would do is putting `if (dateTime is  close to the time now)` in `android alarm`.dart, so play alarm and not call `AndroidAlarmManager.oneShotAt`

**Screenshots**
![image](https://github.com/gdelataillade/alarm/assets/63166757/66f906d9-5aec-4ff1-9a8c-e98b7f98c518)

**Device info**
Pixel 6, android 13

gdelataillade:
Hi @dtkdt100,

Thanks for your interest in the plugin.

I must use `AndroidAlarmManager.oneShotAt` because if I don't, the alarm won't ring if the app is in background. This method creates an isolate to ensure that the code in `playAlarm` is executed whether app is in foreground or background.

Furthermore, I'm not sure that's the reason of the delay. I think a reason that could explain the delay is if the audio asset is large, because the player has to load it before being able to play it.

Do you have this issue with the example app ? If not, is your audio asset large ?

dtkdt100:
Hi, @gdelataillade . Thank you so much for replaying so fast!

The whole point of my issue that you don't need to use background alarm, because the datetime is so close (I mean in a metter of milliscronds to one second). Checkout my fork https://github.com/Dolev-Franco/alarm It is working good (you just need to add a if statement - `if date is close`).

My asset is not to large, I know it because in my fork it working well.

gdelataillade:
Hi @dtkdt100 

OK I understand your request. I'll add it to the plugin as soon as I find the time. I'll let you know when it's done.

dtkdt100:
@gdelataillade 
I am going to make a PR for this

dtkdt100:
@gdelataillade 

[...]

Issue #37, opened 2023-04-13
[Android] Alarm sound is dependent on the "Media volume" but there is a seperate volume called "Alarm volume"


dtkdt100:
Alarm sound is dependent on the "Media volume" but there is a seperate volume called "Alarm volume", sometimes there is a need to play alarm with sound even when the media volume is down, like native alarm clock.

Checkout: https://github.com/ryanheise/audio_session.

Please make it an option and it will almost like native alarm clock!

It's not hard. Just add this:

```dart
final session = await AudioSession.instance;
    await session.configure(const AudioSessionConfiguration(
      androidAudioAttributes: AndroidAudioAttributes(
        usage: AndroidAudioUsage.alarm,
      ),
    ));
```

gdelataillade:
Good point. I could add a `maxVolume` optional parameter. Thanks !

gdelataillade:
Hey @dtkdt100

I was working on this feature and I realized that due to privacy reasons, there is no official way to change device alarm volume in iOS. I really want the `alarm` plugin to have the exact same behavior for Android and iOS so I will probably not add this feature.

I am adding this to the README FAQ:

> ### Why isn't my alarm ringing when the device volume is off?
> 
> iOS prevents third-party apps from modifying volumes. If media volume is off, then the alarm is muted. To ensure the alarm rings, users should turn on media volume before setting the alarm. For consistency, we don't add volume checks for Android, maintaining a uniform user experience on both platforms. Feel free to add one in your app is desired.

Let me know if you have a comment or any other question.

dtkdt100:
@gdelataillade I think in order to achive this behavoir, alarm apps in iOS are turning up the volume and then turring it down again. (https://pub.dev/packages/perfect_volume_control)

Furthermore, you could use these parmas to stop and replay any other music that the iphone user:

```dart
avAudioSessionCategory: AVAudioSessionCategory.playback,
avAudioSessionCategoryOptions: AVAudioSessionCategoryOptions.none,
```

in the audio_session package^.

Maybe you could to the same behavoir with that?

gdelataillade:
I think the `perfect_volume_control` changes the volume relatively to the device volume set by the user. I mean if the device volume is 70% and I set it to 1.0, then my audio will player with volume 70%. If I set it to 0.5, then it will play with 35%.

Good point for the `audio_session` feature thanks.

dtkdt100:
[...]

PR #68, opened 2023-07-21
[Android] Fix: Alarm rings NOW in DateTime.now (in a difference of 1 sec)

dtkdt100:
FIx (#66)

Before:

https://github.com/gdelataillade/alarm/assets/63166757/aabe131c-0b7d-4121-9f90-6f8cd0044957

After

https://github.com/gdelataillade/alarm/assets/63166757/53cc2e48-dfc0-4efa-ae4b-88e21a932e5d


gdelataillade:
Hi @dtkdt100,

Your added code will only run when the alarm is set to sound at the precise moment it is activated. Is this what you want ? What's the use case ?

dtkdt100:
@gdelataillade For me the use case is:
I am developing an app that do alarm by GPS and when the user will arrive at the right location I want the alarm to ring now.

Also, in your example app there is a button for "Ring now", so I tought if it is in your example app it must be used for something

gdelataillade:
Hi @dtkdt100,

OK I see. The NOW button in the example is to show the proof-of-concept.

gdelataillade:
Did you make some tests ? Your code works good ?

dtkdt100:
@gdelataillade Yes.

Tests:
1. Rings now (as you can see in the video)
2. Set an alarm for one min later (works find, alarm and notificaiton are working good)

gdelataillade:
Does your GPS app has to ring even if the app is on background ?

dtkdt100:
Indeed.
I am using `fullScreenIntent: true` in `FlutterLocalNotificaion` and `AudioSession` in Android or `VolumeController` in iOS in oder to set the volume up as suggested in https://github.com/gdelataillade/alarm/issues/37

gdelataillade:
With the code you added, are you sure the alarm will ring in the background ? Because the point of using `AndroidAlarmManager` is to be able to run the Flutter code even if the app is on background.

dtkdt100:
Yes it is working for me. But i am not sure if it is because of `fullScreenIntent: true`.

But in this case there is not option that the phone will be in background because when you start the alarm the app will be in foreground, the alarm rings immediately

gdelataillade:
OK cool, then give me a few days to run some tests and take a look at `fullScreenIntent: true`. I'll keep you updated. In the meantime you can use your fork for your project.

Thanks for your work.

dtkdt100:
Thanks for your quick response!

lib/src/android_alarm.dart
```diff
@@ -67,6 +67,17 @@ class AndroidAlarm {
       throw AlarmException('Isolate error: $e');
     }
 
+    if (dateTime.difference(DateTime.now()).inSeconds <= 1) {
+      await playAlarm(id, {
+        "assetAudioPath": assetAudioPath,
+        "loopAudio": loopAudio,
+        "vibrate": vibrate,
+        "fadeDuration": fadeDuration.toInt(),
+        "enableNotificationOnKill": enableNotificationOnKill,
```
gdelataillade, line 76: We don't need to pass the arguments `vibrate` and `enableNotificationOnKill` right ?

gdelataillade (changes requested):

lib/src/android_alarm.dart
gdelataillade: Any update ? Otherwise I can do it myself if you want.

dtkdt100: Oops, didn't notice your comment. 
I am on a trip right now, so I can't do it now.
I will arrive in three days and fix it. It is an easy fix you could do it yourself if you want.

gdelataillade: It's your PR so I can't modify it directly, but no worries I'll wait !

dtkdt100: Thanks!

dtkdt100:
@gdelataillade 
Ready to merge

gdelataillade (approved):
OK

gdelataillade:
Merged to main branch ! Thanks for your contribution !

PR #70, opened 2023-07-27
Feature -> Add volumeMax parameter

gdelataillade:
Add a boolean parameter to set the system volume to maximum to make sure the user will hear the alarm ring.


gdelataillade:
Feature added in #72

PR #72, opened 2023-08-01
Feature -> [iOS] Keep app active on background

gdelataillade:
- [x] [iOS] Play looping silent audio player to keep app active even when backgrounded to be able to execute code instantly when alarm starts ringing.
- [x] Add a volumeMax parameter.
- [x] Add documentation.
- [x] Update example app.
- [x] Add volumeMax & vibrations in alarm behavior table in the README.


gdelataillade (commented):
OK

PR #75, opened 2023-08-08
Make iOS work with absolute file paths.

henry-lang:
My tech team encountered an error in our app where we use flutter_cache_manager in conjunction with this library, because your swift code currently doesn't work with absolute file paths which is what a File.path returns for us. I have currently modified the code to always just do that, but a feature toggling the type of path used through an argument could be nice. Please consider this feature as it drastically expanded the library's abilities since it can work with dynamic data not stored in the bundle.

Sincerely,
Henry Langmack

### DIFF
--- a/example/lib/screens/home.dart
+++ b/example/lib/screens/home.dart
@@ -98,7 +98,7 @@ class _ExampleAlarmHomeScreenState extends State<ExampleAlarmHomeScreen> {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      appBar: AppBar(title: const Text('alarm 3.0.1')),
+      appBar: AppBar(title: const Text('alarm 3.0.2')),
       body: SafeArea(
         child: alarms.isNotEmpty
             ? ListView.separated(


gdelataillade:
Hey @henry-lang,

Thanks for the PR on the plugin! I'll check it out soon. In the meantime, you should use it directly in your project. Just wondering, have you tested it thoroughly? How's it on Android?

Best,

henry-lang:
I can do some testing if you would like, as I would love to see this feature get merged into the library

gdelataillade:
Hi @henry-lang

Sorry I've been very busy lately, working on another PR. I'd love to see this feature merged too !

I you could do some testing and send me what you tested and the results, it would really help. Keep in mind that the plugin has to keep the exact same behavior on iOS and Android.

Thank you in advance !

henry-lang:
For sure, I’ll get back to you later this week

gdelataillade:
Hey @henry-lang any update ?

henry-lang:
Sorry, been busy, testing shouldn't be too bad, we're running a production app with it and it's fine

PR #125, opened 2023-12-25
Fixed pickTime

HARUM1122:
Changes Made
Fixed a problem in the pickTime function where setting an alarm for today would sometimes set it for tomorrow. This happened when choosing a time later than the current time. The issue was caused by mistakenly using .copyWith with selectedDateTime, which already had an extra day. I updated the logic to use DateTime.now().copyWith instead for accurate behavior in both today and tomorrow scenarios.

Problem
Previously, when users set an alarm for today and picked a later time, the alarm was mistakenly scheduled for tomorrow. This occurred when the selected time was after the current time.

Solution
Changed selectedDateTime = selectedDateTime.copyWith to selectedDateTime = DateTime.now().copyWith to fix the issue. Now, when users set alarms for today, it works correctly without adding an extra day.

I've tested it out and now it's working fine. I've also replaced if-elif-else with switch statements.

### DIFF
--- a/example/lib/screens/edit_alarm.dart
+++ b/example/lib/screens/edit_alarm.dart
@@ -47,14 +47,15 @@ class _ExampleAlarmEditScreenState extends State<ExampleAlarmEditScreen> {
     final today = DateTime(now.year, now.month, now.day);
     final difference = selectedDateTime.difference(today).inDays;
 
-    if (difference == 0) {
-      return 'Today';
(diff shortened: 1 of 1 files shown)


gdelataillade (approved):

gdelataillade:
Hi @HARUM1122 

Thanks for your contribution ! I really appreciate !

PR #145, opened 2024-02-01
Feature/prevents int type error

Raonshi:
## Why this PullRequest is needs
When i make `AlarmSettings` model, I set `id` to timestamp of DateTime.now(). It works on iOS, but doesn't work on Android.
So, to prevent this case, I added some assertions in `AlarmSettings` model.

## Changes
- Add  two assertion in `AlarmSettings` model.
  - `assert(id != 0 && id != -1, "id cannot be set to 0 or -1")`
  - `assert(id <= 2147483647, "id cannot be set larger then Android Int.MAX_VALUE")`
- Add example case

### DIFF
--- a/example/lib/screens/edit_alarm.dart
+++ b/example/lib/screens/edit_alarm.dart
@@ -69,12 +69,11 @@ class _ExampleAlarmEditScreenState extends State<ExampleAlarmEditScreen> {
       setState(() {
         final DateTime now = DateTime.now();
         selectedDateTime = now.copyWith(
-          hour: res.hour,
-          minute: res.minute,
-          second: 0,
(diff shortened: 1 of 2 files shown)


gdelataillade:
Hi @Raonshi 

Thank you so much for your PR !

Sounds good, let me know when you're done and I'll review it and merge it.

Raonshi:
@gdelataillade 
actually, i'm already done that.
in this PR, you can see two assertions in AlarmSettings model.

this PR supports developer to prevent using big number in id of AlarmSettings at debug mode.

gdelataillade (commented):
OK

Issue #146, opened 2024-02-03
Alarm id issue


rajnishsharma1:
**Alarm plugin version**
alarm: ^3.0.7

**Describe the bug**
```
E/flutter (23387): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Alarm id cannot be set smaller than Int min value (-2147483648). Provided: 42
E/flutter (23387): #0      Alarm.alarmSettingsValidation (package:alarm/alarm.dart:104:7)
```

**To Reproduce**
*Steps to reproduce the behavior:*
1. Try setting up an alarm
2. Code:
```
    final alarmSettings = AlarmSettings(
      id: 42,
      dateTime: dateTime,
      assetAudioPath: 'assets/alarm.mp3',
      loopAudio: true,
      vibrate: true,
      volume: 0.8,
      fadeDuration: 3.0,
      notificationTitle: 'This is the title',
      notificationBody: 'This is the body',
      enableNotificationOnKill: true,
    );

    await Alarm.set(alarmSettings: alarmSettings);
```

**Device info**
Vivo, Android 12

tattuu:
same issue here

rajnishsharma1:
@tattuu It seems this issue is introduced in newer version
alarm: ^1.2.1 is not having this issue.

pyjserv:
I encountered the same issue. It was fine in version 3.0.6.

JosipDomazetDev:
This pull request should resolve this issue,  a "-" was missing.
https://github.com/gdelataillade/alarm/pull/148

gdelataillade:
Hi @rajnishsharma1 @tattuu @pyjserv @JosipDomazetDev 

Really sorry about that, it was a silly issue. Thanks @JosipDomazetDev for your PR.

I just released version `3.0.8` with the fix.

JosipDomazetDev:
Thanks for the merge, i just wanna mention it was the PR of @Rest00rer not mine, I just commented on both threads to make sure you see it as soon as possible ^^

gdelataillade:
Alright, then thank you @Rest00rer then !

PR #148, opened 2024-02-03
Fix alarm id verification

Rest00rer:
Fix Exception 'Alarm id cannot be set smaller than Int min value (-2147483648). Provided: ${alarmSettings.id}'


JosipDomazetDev:
This should resolve the following issue https://github.com/gdelataillade/alarm/issues/146

PR #246, opened 2024-10-02
Fix fatal crash on iOS due to incorrect arg names

orkun1675:
Realized this when trying to run my app on iOS after upgrading to alarm to `4.0.0`.


gdelataillade (approved):
LGTM

gdelataillade:
Thanks for the PR !

gdelataillade:
Added to `4.0.1`

Issue #248, opened 2024-10-06
Android crashes on v4.0.1


aykutuludag:
**Alarm plugin version**
*v4.0.1*

**Describe the bug**
*Some crash reports from Firebase Crashlytics*

**com.gdelataillade.alarm.alarm.AlarmService.onStartCommand
java.lang.NullPointerException - Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference**

```
Fatal Exception: java.lang.RuntimeException: Unable to start service com.gdelataillade.alarm.alarm.AlarmService@386a492 with Intent { cmp=PACKAGE_NAME/com.gdelataillade.alarm.alarm.AlarmService (has extras) }: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
       at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:4156)
       at android.app.ActivityThread.access$1900(ActivityThread.java:228)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1923)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7438)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:989)
```

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
       at org.json.JSONTokener.nextCleanInternal(JSONTokener.java:121)
       at org.json.JSONTokener.nextValue(JSONTokener.java:98)
       at org.json.JSONObject.<init>(JSONObject.java:164)
       at org.json.JSONObject.<init>(JSONObject.java:181)
       at com.gdelataillade.alarm.alarm.AlarmService.onStartCommand(:149)
       at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:4138)
       at android.app.ActivityThread.access$1900(ActivityThread.java:228)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1923)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7438)
[...]

PR #249, opened 2024-10-06
Fix -> Android native exceptions

aykutuludag:
com.gdelataillade.alarm.alarm.AlarmService.onStartCommand java.lang.NullPointerException - Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference

### DIFF
--- a/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
@@ -67,13 +67,17 @@ class AlarmService : Service() {
         val fadeDuration = intent.getDoubleExtra("fadeDuration", 0.0)
         val fullScreenIntent = intent.getBooleanExtra("fullScreenIntent", true)
 
+        // Fetching notificationSettings safely or using default
         val notificationSettingsJson = intent.getStringExtra("notificationSettings")
-        val jsonObject = JSONObject(notificationSettingsJson)
-        val map: MutableMap<String, Any> = mutableMapOf()
-        jsonObject.keys().forEach { key ->
-            map[key] = jsonObject.get(key)
+        val notificationSettings = if (!notificationSettingsJson.isNullOrEmpty()) {
+            val jsonObject = JSONObject(notificationSettingsJson)
+            val map: MutableMap<String, Any> = mutableMapOf()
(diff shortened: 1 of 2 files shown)


aykutuludag:
https://github.com/gdelataillade/alarm/issues/248

android/src/main/kotlin/com/gdelataillade/alarm/alarm/BootReceiver.kt
```diff
@@ -33,6 +33,8 @@ class BootReceiver : BroadcastReceiver() {
 
         for (alarm in storedAlarms) {
             // Create the arguments for the MethodCall
+            val notificationSettings = alarm.notificationSettings ?: NotificationSettings.default()
+
```
gdelataillade, line 37: It looks like you added this new variable but you're not using it.

android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
```diff
@@ -67,13 +67,17 @@ class AlarmService : Service() {
...
-        val jsonObject = JSONObject(notificationSettingsJson)
-        val map: MutableMap<String, Any> = mutableMapOf()
-        jsonObject.keys().forEach { key ->
-            map[key] = jsonObject.get(key)
+        val notificationSettings = if (!notificationSettingsJson.isNullOrEmpty()) {
+            val jsonObject = JSONObject(notificationSettingsJson)
+            val map: MutableMap<String, Any> = mutableMapOf()
+            jsonObject.keys().forEach { key -> map[key] = jsonObject.get(key) }
+            NotificationSettings.fromJson(map)
+        } else {
```
gdelataillade, line 77: Maybe remove the else block ? Well notificationSettingsJson is never supposed to be null or empty but if it's the case then we don't want to show a notification.

aykutuludag: It is empty when users switched from old alarm library versions if I understand correctly.

gdelataillade: you're right. I realize I have to add backward compatibility with older models in native code. Don't worry I'll add it myself. You can close this PR. Thanks for pointing this issue out.

PR #250, opened 2024-10-07
Fix bootreceiver bug (android)

aykutuludag:


gdelataillade:
The issue was a lack of backward compatibility with older AlarmSettings model when device rebooted. I added this code to fix it, in `rescheduleAlarms`:


```Kotlin
if (alarm.notificationSettings == null) {
        // Create notificationSettings from notificationTitle and notificationBody
        alarm.notificationSettings = NotificationSettings(
                    title = alarm.notificationTitle ?: "Title",
                    body = alarm.notificationBody ?: "Body"
         )
}

if (alarm.enableNotificationOnKill == null) {
       alarm.enableNotificationOnKill = alarm.warningNotificationOnKill ?: false
}
```

I'm publishing this fix in release `4.0.3` in the next minutes.

Issue #267, opened 2024-10-24
iOS: Alarm volume does not fade


orkun1675:
**Alarm plugin version**
4.0.7

**Describe the bug**
Instead of slowly fading to full volume, the alarm starts ringing at the full volume.

**To Reproduce**
1. We use an audio file that is silent for 5 seconds and loud for 30+ seconds
2. When creating alarm set the following parameters:

```dart
loopAudio: true,
vibrate: true,
volume: 0.8,
fadeDuration: 12, // seconds
```

3. Wait for alarm to ring
4. Observe how the volume is already at 80% at the 5 second mark

**Expected behavior**
The volum should increase linearly over 12 seconds. So we would expect to hear 5/12*0.8 = 33% volume alarm at the 5 second mark.

**Screenshots**
N/A

**Device info**
iPhone SE, iOS 18.0.1, physical device

**Additional context**
Note that this is WAI on Android.

gdelataillade:
I’ve just released version `4.0.8` with a fix. Please let me know if it resolves the issue. Thanks!

orkun1675:
Even with a fade duration of 30 seconds I wasn't able to verify this is working, it still feels like the volume is constant.

gdelataillade:
Are you sure the volume is constant ? If you want you can build the app with Xcode, you will see more logs like this:  


![CleanShot 2024-10-29 at 16 34 18@2x](https://github.com/user-attachments/assets/e61317c4-7af5-4671-b1de-41ddf145fb45)

If you want to check out the code, take a look at the [SwiftAlarmPlugin.fadeVolume](https://github.com/gdelataillade/alarm/blob/4bb398f5a0d68b85c62f6970bbe73095b9697496/ios/Classes/SwiftAlarmPlugin.swift#L300) method.

PR #276, opened 2024-11-01
[Feature] Allow specifying the alarm volume using steps/stops

orkun1675:
This PR allows `alarm` users to speicfy a staircase function to control the alarm volume.

It also fixes https://github.com/gdelataillade/alarm/issues/267 by adding [this line](https://github.com/gdelataillade/alarm/pull/276/files#diff-a6e190b4d0361c3488dc1010094bb54270f9ff9f6ab56bb22692e17fff612945R143). For some reason AVAudioPlayer ignores volume changes unless the initial volume is set to zero.

PS: Overall, AVAudioPlayer has been a pain to deal with, `setVolume` has bugs, and it also doesn't report the real `.volume` value while a fade is in progress.

I've also taken the liberty to fix lint warnings in both iOS and Android native code.

I've tested using iOS and Android emulators + iOS physical device. I would appreciate if you could test this as well before merging. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.0.10
+* [Android] Fixed `androidFullScreenIntent` parameter.
+
 ## 4.0.9
 * Added new alarm parameter `volumeEnforced`.
 * [iOS] Reopen killed app when tap on notification stop button.
--- a/example/lib/screens/home.dart
+++ b/example/lib/screens/home.dart
@@ -9,7 +9,7 @@ import 'package:alarm_example/widgets/tile.dart';
 import 'package:flutter/material.dart';
 import 'package:url_launcher/url_launcher.dart';
 
-const version = '4.0.9';
+const version = '4.0.10';
 
 class ExampleAlarmHomeScreen extends StatefulWidget {
   const ExampleAlarmHomeScreen({super.key});
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -7,7 +7,7 @@ packages:
       path: ".."
       relative: true
     source: path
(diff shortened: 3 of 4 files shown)


gdelataillade (approved):

gdelataillade:
Thanks for the PR! I have one remark: it might be a bit much to have three separate parameters just for handling volume fade. Maybe we could create a new model, like I did with `NotificationSettings`, to group related parameters under the same theme. For example, the `volume` parameter and the new `volumeEnforced` parameter could be part of a model called `VolumeSettings`. I realize it would take some extra work to refactor, but what do you think of this approach?

orkun1675:
Hi @gdelataillade I totally agree the current API is not ideal. It also relies on the caller providing two lists of equal length + the first list having strictly inceasing float values. It would be much better to communicate these requirements with an object oriented interface.

However, I would prefer to decouple these improvements from this feature PR. It would be much easier to implement this API refactoring after migrating to [pigeon](https://pub.dev/packages/pigeon), WDYT?

gdelataillade:
Got it, so your plan is:

1) Merge this PR
2) Migrate to pigeon
3) Refactor to an object-oriented API

A couple of thoughts: Wouldn’t adding these parameters now be confusing for users? Also, migrating to Pigeon could take a while, and I’m not sure it’s needed since the current setup already has models and native data handling working well. I haven’t used Pigeon myself, so I’m not sure how well it would fit here. Are you comfortable handling the migration if we go that route?

orkun1675:
I'll see if I can carve out sometime to look into Pigeon migration. It's OK by me to keep this PR pending in the meantime,

PR #282, opened 2024-11-11
[Feature] Allow specifying the alarm volume using steps/stops

orkun1675:
(updated version of https://github.com/gdelataillade/alarm/pull/276)

This PR allows alarm users to speicfy a staircase function to control the alarm volume.

It also fixes https://github.com/gdelataillade/alarm/issues/267 by adding [this line](https://github.com/gdelataillade/alarm/pull/276/files#diff-a6e190b4d0361c3488dc1010094bb54270f9ff9f6ab56bb22692e17fff612945R143). For some reason AVAudioPlayer ignores volume changes unless the initial volume is set to zero.

PS: Overall, AVAudioPlayer has been a pain to deal with, setVolume has bugs, and it also doesn't report the real .volume value while a fade is in progress.

I've also taken the liberty to fix lint warnings in both iOS and Android native code.

I've tested using iOS and Android emulators + iOS physical device. I would appreciate if you could test this as well before merging. Thanks!


gdelataillade:
Hi @orkun1675 

Thanks for the PR ! Sure, I will test this on my devices. Are we doing the pigeon migration in this PR ?

orkun1675:
This PR is a clone of the previous one (when I moved my changes from `main` to a new branch GitHub closed to previous PR).

PS: Today I started the Pigeon migration, Dart + Kotlin refactorings are complete, Swift is pending.

orkun1675:
@gdelataillade this PR is now rebased to main and updated to utilize Pigeon and a more friendly public API. PTAL.

CHANGELOG.md
gdelataillade, line 1: The version should be updated to `5.0.0`, not `4.2.0`, because renaming API parameters introduces breaking changes that require user refactoring, even if no features are removed.

Many developers use the caret constraint (^) in their `pubspec.yaml`, so releasing this as `4.2.0` could unintentionally upgrade their package during `flutter pub get` and introduce breaking changes. Incrementing to `5.0.0` ensures users explicitly opt-in for the update. What do you think?

Source: https://dart.dev/tools/pub/versioning

gdelataillade (changes requested):

CHANGELOG.md
orkun1675: Makes sense, updated.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 5.0.0
+* **BREAKING**: Old alarms (alarms created pre v5) will be deleted.
```
gdelataillade, line 2: Is there a way to add compatibility with pre v5 ? Like I did in v4 when I added `NotificationSettings` model to `AlarmSettings`. Because otherwise users of all the apps using alarm package will loose their alarms and there's no way to notify them... What do you think ? I can work on that if you want.

gdelataillade (changes requested):

CHANGELOG.md
orkun1675: We might need to do this in multiple languages. Also it's the serialization logic between Kotlin and Dart coupled? It would be great to decouple them: Kotlin saves its own copy of alarms in local storage.

It would be great if you have time to look at this!

example/android/app/build.gradle
```diff
@@ -25,7 +25,7 @@ if (flutterVersionName == null) {
 android {
     namespace 'com.gdelataillade.alarm.alarm_example'
 
-    compileSdkVersion 34
+    compileSdkVersion 35
```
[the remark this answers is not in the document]
orkun1675: I don't think so, this was mostly for me since I run a SDK 35 Android pixel emulator.

example/lib/screens/edit_alarm.dart
```diff
@@ -29,19 +31,20 @@ class _ExampleAlarmEditScreenState extends State<ExampleAlarmEditScreen> {
     creating = widget.alarmSettings == null;
 
     if (creating) {
-      selectedDateTime = DateTime.now().add(const Duration(minutes: 1));
-      selectedDateTime = selectedDateTime.copyWith(second: 0, millisecond: 0);
+      selectedDateTime = DateTime.now().add(const Duration(seconds: 10));
```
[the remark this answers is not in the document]
orkun1675: Good catch, I can revert this when I get access to a laptop. 

I'd also consider keeping it as it greatly speeds up testing.

orkun1675:
@gdelataillade can we merge this so I can rebase and work on my follow-up PRs?

We don't need to release a new version just yet if you'd like to work on the backwards compatibility aspect.

gdelataillade:
@orkun1675  Yes, I intend to ensure backward compatibility before releasing version `5.0.0`. However, I’m uncertain when I’ll have the time to complete this.

In the meantime, I can provide a development release, version `5.0.0-dev.1`, if you’d like to incorporate it into your project.

onegigbyte:
# Volume fade still broken in iOS (v5.0.0-dev.1)

Hey, I just wanted to report that the iOS volume fade issue from #267 is still happening in 5.0.0-dev.1. I tried setting a 30-second fade, but the volume still jumps to max in like 2-3 seconds.

## What I tried
I tested with the latest dev version (5.0.0-dev.1) using both the regular fade and staircase fade:
```dart
final VolumeSettings volumeSettings = VolumeSettings.fade(
    volume: 1.0,
    fadeDuration: Duration(seconds: 100),
);
```

## What's happening
When I run this, I see the log message:
"[SwiftAlarmPlugin] Fading volume to 1.0 over 99.95 seconds."

But in reality, the volume maxes out after just 2-3 seconds.

## I think I found the problem
Looking at the code in [ios/Classes/api/AlarmApiImpl.swift (L397-403)](https://github.com/gdelataillade/alarm/blob/main/ios/Classes/api/AlarmApiImpl.swift#L397-L403):

```swift
DispatchQueue.main.asyncAfter(deadline: now + startTime) {
    if !audioPlayer.isPlaying {
        return
    }
    print("[SwiftAlarmPlugin] Fading volume to \(targetVolume) over \(fadeDuration) seconds.")
    audioPlayer.setVolume(targetVolume, fadeDuration: fadeDuration)
}
```

I think there's a timing issue here. 
- The code is adding seconds directly to `DispatchTime.now()`, 
- but `DispatchTime` actually works in nanoseconds. 
This is likely to explain why everything's happening way faster than it should.

## Suggested fix
We probably need to convert the seconds to nanoseconds before adding them to the DispatchTime:
```swift
let delayInNanoseconds = UInt64(startTime * Double(NSEC_PER_SEC))
DispatchQueue.main.asyncAfter(deadline: now + .nanoseconds(Int(delayInNanoseconds)))
```

gdelataillade:
@onegigbyte a fix for your issue was published in `5.0.0-dev.5`.
Thanks @orkun1675 !

Issue #265, opened 2024-10-24
iOS: Alarm.isRinging(alarmId) is FALSE after alarm rings


orkun1675:
**Alarm plugin version**
4.0.7

**Describe the bug**
After the alarm rings calling `Alarm.isRinging(alarmId)` returns false. This happens about 30% of the time.

**To Reproduce**
1. Set an alarm
2. Listen for `Alarm.ringStream.stream`
3. When an event fires in this stream, call Alarm.isRinging(alarmId)`
4. PS: At this point our code schedules a timer to repeat 3 every 1s to figure out when the alarm is stopped by the user

**Expected behavior**
`Alarm.isRinging(alarmId)` should continue to return to until the alarm is stopped. 

**Screenshots**
N/A

**Device info**
iPhone SE, iOS 18.0.1, physical device

**Additional context**
Note that the `Alarm.updateStream.stream` is also buggy and can't be used right now to identify when active alarms are stopped. For instance; [the set function](https://github.com/gdelataillade/alarm/blob/main/lib/alarm.dart#L73) returns too early and never populates the stream on line 88.

gdelataillade:
I’ve just released version `4.0.8` with a fix. Please let me know if it resolves the issue. Thanks!

gdelataillade:
By the way, `id` is now optional:

```Dart
Future<bool> isRinging([int? id])
```

> If no `id` is provided, it checks if any alarm is ringing.
> If an `id` is provided, it checks if the specific alarm with that `id`

gdelataillade:
I'm not sure to understand the issue. Is it because `Alarm.isRinging` returns `false` immediately after calling `Alarm.set` ?

orkun1675:
>  Is it because Alarm.isRinging returns false immediately after calling Alarm.set ?

Yes, exactly.

After working on a PR I have seen that we are using `AVAudioPlayer.isPlaying` to check if an alarm is active. However, doesn't this return `false` while the audio file is being loaded by `AVAudioPlayer. prepareToPlay`? Maybe this is causing a race condition?

PR #283, opened 2024-11-11
[iOS] Use triggerTime to determine if alarm is ringing

orkun1675:
Fixes https://github.com/gdelataillade/alarm/issues/265


ios/Classes/SwiftAlarmPlugin.swift
```diff
@@ -55,7 +55,7 @@ public class SwiftAlarmPlugin: NSObject, FlutterPlugin {
             if id == nil {
                 result(self.isAnyAlarmRinging())
             } else {
-                result(self.alarms[id!]?.audioPlayer?.isPlaying ?? false)
+                result(self.alarms[id!]?.triggerTime?.timeIntervalSinceNow ?? 1 <= 0)
```
gdelataillade, line 58: I tested `isRinging` and noticed that it returns false the first time a build session is started, but consistently returns true in subsequent checks. This behavior seems tied to how `timeIntervalSinceNow` evaluates, as I observed it returning values around `0.25`.

Do you encounter this issue as well? If so, it might be worth considering a threshold adjustment. Instead of checking with `<= 0`, perhaps using `<= 0.5` could account for this variation. What are your thoughts on adjusting this tolerance to handle edge cases more robustly?

ios/Classes/SwiftAlarmPlugin.swift
```diff
@@ -55,7 +55,7 @@ public class SwiftAlarmPlugin: NSObject, FlutterPlugin {
             if id == nil {
```
gdelataillade, line 55: I made a mistake here. `id` is not equal to `call.arguments`. It has to be extracted from a map, like in `stopAlarm` above:

```Swift
if let arguments = call.arguments as? [String: Any],
    let id = arguments["id"] as? Int {
        // id is specified
} else {
        // id is not specified
}
```

gdelataillade (changes requested):

ios/Classes/SwiftAlarmPlugin.swift
gdelataillade: I just released the hotfix in `4.0.11`.

ios/Classes/SwiftAlarmPlugin.swift
orkun1675: I was able to replicate this issue.

This was hapenning because the Flutter code was using `AlarmSettings.dateTime` to evaluate `isRinging` and adding the alarm to the `ringStream`.

In the meantime, the iOS code was using `audioPlayer.isPlaying` OR `alarms[id].triggerTime` to evaluate `isRinging`. Both of these variables are subject to the audio asset load delay.

I've resolved this by making a call from iOS --> Flutter when the alarm rings/stops; so that platform code is the source of truth. This means I was also able to delete the periodic checks in the `ios_alarm.dart` file.

I've tested in multiple scenarious and this is WAI now.

ios/Classes/SwiftAlarmPlugin.swift
orkun1675: Thanks!

orkun1675:
Please see the latest commit (https://github.com/gdelataillade/alarm/pull/283/commits/7caa0211005cd99ac57481c5c9c755178d5cae40) for the diffs.

PR #292, opened 2024-11-24
Migrate to Pigeon

orkun1675:
This PR refactors Flutter-Platform communications using Pigeon.

It also simplifies some code, improves error handling, updates dependencies, and fixes some lint warnings.


android/src/main/kotlin/com/gdelataillade/alarm/alarm/BootReceiver.kt
```diff
@@ -28,62 +19,20 @@ class BootReceiver : BroadcastReceiver() {
     private fun rescheduleAlarms(context: Context) {
```
gdelataillade, line 19: I ran a few tests on my Android device and it looks like alarms are not being rescheduled after device reboot.

lib/alarm.dart
```diff
@@ -175,12 +194,14 @@ class Alarm {
...
   /// Returns all the alarms.
-  static List<AlarmSettings> getAlarms() => AlarmStorage.getSavedAlarms();
+  static Future<List<AlarmSettings>> getAlarms() =>
+      AlarmStorage.getSavedAlarms();
 
   /// Reloads the shared preferences instance in the case modifications
   /// were made in the native code, after a notification action.
   static Future<void> reload(int id) async {
-    await AlarmStorage.prefs.reload();
+    // TODO(orkun1675): Remove this function and publish stream updates for
```
gdelataillade, line 203: TODO

pigeons/alarm_api.dart
```diff
@@ -0,0 +1,212 @@
...
+
+  void setWarningNotificationOnKill({
+    required String title,
+    required String body,
+  });
+
+  void disableWarningNotificationOnKill();
+}
+
+// TODO(orkun1675): Consider migrating to Pigeon type-safe channels when they
```
gdelataillade, line 203: TODO

ios/Classes/services/NotificationManager.swift
gdelataillade, line 1: I noticed a change where triggering STOP_ACTION to stop the alarm now also opens the app. It seems like in the current version, the app only opens if it was previously killed. Was this behavior intentional, or could it be an unintended change? Thanks for clarifying!

gdelataillade (changes requested):
Thank you for all the hard work! I’ve left a few comments on your changes—please feel free to reach out if you need any help or have any questions. Great job!

lib/service/alarm_storage.dart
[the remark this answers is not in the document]
orkun1675: The SharedPreferencesAsync initializer does not do any work right now so this is not required, but is not a bad idea to future proof this. Done!

example/lib/screens/home.dart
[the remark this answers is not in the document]
orkun1675: Sorry about that, the fix for this was in https://github.com/gdelataillade/alarm/pull/294 which I've now moved to this PR.

pigeons/alarm_api.dart
orkun1675: Removed, there is no reason to use streams to support this API signature.

lib/alarm.dart
orkun1675: This is out of scope of this migration PR. I added a TODO here because updating the `ringStream` when an alarm stops ringing would be great feature addition and should be easy to do in a follow-up PR.

android/src/main/kotlin/com/gdelataillade/alarm/alarm/BootReceiver.kt
orkun1675: This is the console logs after device reboot on Pixel 8 API 35:

```
2024-11-26 21:55:16.607  3165-3165  BootReceiver            com...ataillade.alarm.alarm_example  I  Device rebooted: action is: android.intent.action.BOOT_COMPLETED
2024-11-26 21:55:16.607  3165-3165  BootReceiver            com...ataillade.alarm.alarm_example  D  Device rebooted, rescheduling alarms
2024-11-26 21:55:16.618  3165-3165  BootReceiver            com...ataillade.alarm.alarm_example  D  Rescheduling 0 alarms
```

As you can see the BroadcastReceiver is running but no alarms are found in storage.

This is likely because the `AlarmStorage.saveAlarm` function is never called. Ever before the migration.

I've added a TODO for you so we don't forget to address this bug.

ios/Classes/services/NotificationManager.swift
orkun1675: I've tested this before and after the migration, in both cases the app comes into the foreground for me.

Can you double check if you are experiencing something different with a build from pre-migration?

orkun1675:
All issues addressed, please take another look.

android/src/main/kotlin/com/gdelataillade/alarm/services/AlarmStorage.kt
gdelataillade, line 1: Thank you for your work on the plugin! The reason alarms are not rescheduling after a reboot is due to the migration in `lib/service/alarm_storage.dart` from `shared_preferences`‘s `SharedPreferences` to `SharedPreferencesAsync`. This change affects the ability to access saved alarms from the Android native side, preventing them from being rescheduled by the `BootReceiver`. To fix this, a native migration from Android’s `SharedPreferences` to `DataStore Preferences` is required, as [shown here](https://pub.dev/packages/shared_preferences#storage-location-by-platform).

If you’d like, I can handle this migration for you—just let me know!

gdelataillade (changes requested):

android/src/main/kotlin/com/gdelataillade/alarm/services/AlarmStorage.kt
orkun1675: Unfortunately DataStore interface does not allow access from two different callsites:

`E/flutter ( 7248): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(IllegalStateException, java.lang.IllegalStateException: There are multiple DataStores active for the same file: /data/user/0/com.gdelataillade.alarm.alarm_example/files/datastore/FlutterSharedPreferences.preferences_pb. You should either maintain your DataStore as a singleton or confirm that there is no two DataStore's active on the same file (by confirming that the scope is cancelled)., Cause: null, Stacktrace: java.lang.IllegalStateException: There are multiple DataStores active for the same file: /data/user/0/com.gdelataillade.alarm.alarm_example/files/datastore/FlutterSharedPreferences.preferences_pb. You should either maintain your DataStore as a singleton or confirm that there is no two DataStore's active on the same file (by confirming that the scope is cancelled).`

It seems we won't be able to read Flutter prefs from Android. We will need to maintain state seperately on the Android side.

Thus, I'm reverting to `SharedPreferencesAsync` -> `SharedPreferences`.

gdelataillade:
Added to alarm `4.1.0`. Thanks for the contribution !

PR #301, opened 2024-12-10
Implement stopAll natively

orkun1675:
Having native support for the `stopAll` function should increase reliability since it will ensure all alarms saved on the platform side (Swift, Kotlin) are cleaned up.


gdelataillade (approved):
LGTM

PR #309, opened 2024-12-18
[Feature] Add v4 backward compatibility

gdelataillade:
To ensure seamless transition for users upgrading from v4 to v5, the AlarmSettings class has been modified to include backward compatibility. This update addresses potential issues when retrieving alarms created in v4, ensuring that all settings, including dateTime and volumeSettings, are correctly interpreted and converted in v5.

### DIFF
--- a/android/src/main/kotlin/com/gdelataillade/alarm/models/AlarmSettings.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/models/AlarmSettings.kt
@@ -2,12 +2,17 @@ package com.gdelataillade.alarm.models
 
 import com.gdelataillade.alarm.generated.AlarmSettingsWire
 import kotlinx.serialization.KSerializer
-import java.util.Date
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.descriptors.PrimitiveKind
 import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+import java.time.Duration
+import kotlin.time.toKotlinDuration
+import java.util.Date
 
 @Serializable
 data class AlarmSettings(
@@ -21,7 +26,7 @@ data class AlarmSettings(
     val vibrate: Boolean,
     val warningNotificationOnKill: Boolean,
     val androidFullScreenIntent: Boolean,
-    val allowAlarmOverlap: Boolean = false
+    val allowAlarmOverlap: Boolean = false // Defaults to false for backward compatibility
 ) {
     companion object {
         fun fromWire(e: AlarmSettingsWire): AlarmSettings {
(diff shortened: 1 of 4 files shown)


gdelataillade:
Key changes include:

- **VolumeSettings**: Automatically generate `volumeSettings` for v4 alarms, using existing volume-related fields.
- **DateTime Conversion**: Convert `dateTime` from microseconds (v4) to milliseconds (v5) to ensure accurate alarm scheduling.
- **Default Values**: Set default values for fields introduced in v5, such as `allowAlarmOverlap`.

gdelataillade:
Hi @orkun1675,

I wanted to get your thoughts on the code, as you wrote most of the code I modified, and you have more expertise with native iOS and Android code than I do. If it looks good to you, I'll go ahead and merge it so we can release version `5.0.0` soon.

Thanks!

android/src/main/kotlin/com/gdelataillade/alarm/models/AlarmSettings.kt
```diff
@@ -21,7 +26,7 @@ data class AlarmSettings(
     val vibrate: Boolean,
     val warningNotificationOnKill: Boolean,
     val androidFullScreenIntent: Boolean,
-    val allowAlarmOverlap: Boolean = false
+    val allowAlarmOverlap: Boolean = false // Defaults to false for backward compatibility
```
orkun1675, line 29: I also this this is a better default, I wouldn't expect to alarms to ring at the same time as a novice user.

lib/model/alarm_settings.dart
```diff
@@ -24,8 +24,54 @@ class AlarmSettings {
...
 
   /// Constructs an `AlarmSettings` instance from the given JSON data.
-  factory AlarmSettings.fromJson(Map<String, dynamic> json) =>
-      _$AlarmSettingsFromJson(json);
+  ///
+  /// This factory adds backward compatibility for v4 JSON structures
+  /// by detecting the absence of certain fields and adjusting them.
+  factory AlarmSettings.fromJson(Map<String, dynamic> json) {
+    // Handle backward compatibility for v4 JSON structures
+    if (!json.containsKey('volumeSettings')) {
```
orkun1675, line 32: I'd recommend calling the default JSON parser (` _$AlarmSettingsFromJson(json);`) in a try-catch and falling back to V4 parsing if V5 fails.

This shouldn't hurt performance much since we don't expect this conversion to happen often.

lib/model/alarm_settings.dart
```diff
@@ -40,12 +86,16 @@ class AlarmSettings {
...
         androidFullScreenIntent = wire.androidFullScreenIntent,
-        allowAlarmOverlap = false;
+        allowAlarmOverlap = wire.allowAlarmOverlap;
 
-  /// Unique identifier assiocated with the alarm. Cannot be 0 or -1;
+  /// Unique identifier associated with the alarm. Cannot be 0 or -1.
   final int id;
 
   /// Date and time when the alarm will be triggered.
+  @JsonKey(
```
orkun1675, line 95: Why is this needed? IIRC the default behaviour of `JsonSerializable` is identical?

ios/Classes/models/AlarmSettings.swift
```diff
@@ -13,25 +15,62 @@ struct AlarmSettings: Codable {
...
+                volume: volume,
+                fadeDuration: fadeDuration,
+                fadeSteps: [], // No equivalent for fadeSteps in older models
+                volumeEnforced: volumeEnforced
+            )
+        }
+    }
+
+    /// Encode method to support `Encodable` protocol
+    func encode(to encoder: Encoder) throws {
```
orkun1675, line 58: Could you remind me where this is used?

orkun1675 (commented):
Of course, left a few minor comments.

lib/model/alarm_settings.dart
gdelataillade: Added in last commit.

ios/Classes/models/AlarmSettings.swift
orkun1675: SGTM, I'm just confused how the code didn't break at build time before.

lib/model/alarm_settings.dart
orkun1675: For V4 you are defining custom `fromJson` logic in the function above.

If someone upgrades from V4 to V5 we want `toJson` to use the new String format, right?

So I don't think this additional logic is neccesary.

orkun1675 (commented):
Looks good to me! I'd recommend testing this though as it's hard to ensure data type safety by reading the code.

Looking forward to V5!

lib/model/alarm_settings.dart
gdelataillade: You're right, I've done some tests without it and it works fine.

gdelataillade:
PR added to version `5.0.0-dev.4`.

Issue #311, opened 2024-12-26
alarm listener doesn't fire if app is in foreground (Android only)


ApplySci:
**Alarm plugin version**
5.0.0-dev.4 and 4.1.1

**Describe the bug**
If the app is in the foreground when an alarm goes off, I get a notification. But the alarm listener doesn't get triggered. I'm using `Alarm.ringStream.stream.listen(openAlarmPage)`, but that doesn't get triggered when an alarm starts ringing. In the debug console, I do see `Flutter was notified that alarm 1 is ringing`.

**Additional context**
I am using firebase cloud messaging and handlers for notifications from that, but they're not firing when an alarm sounds, either (not that I'd expect them too - I'm just adding this in case there's  some weird collision there)

ApplySci:
NB it works fine on ipad - it's specifically an android issue.

Because of the new Play Store restrictions, I've got this in my `AndroidManifest.xml`:
```
    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM"/>
    <uses-permission android:name="android.permission.USE_EXACT_ALARM" tools:node="remove"
        tools:ignore="ExactAlarm" />
    <uses-permission android:name="android.permission.USE_FULL_SCREEN_INTENT" tools:node="remove" />
```

ApplySci:
In `AlarmPlugin.alarmTriggerApi?.alarmRang(id.toLong())`, in debugging, I get:

`FlutterError: Unable to establish connection on channel: 'dev.flutter.pigeon.alarm.AlarmTriggerApi.alarmRang'.`

gdelataillade:
Hi @ApplySci 

Sorry for the late reply. Do you still have the issue ?

@orkun1675 do you have an idea of what could be causing the issue ?

orkun1675:
@ApplySci I'm happy to look into this. Can you share a minimum reproducible example? For instance, can modify the `example` app to get it to print the error message you mentioned above?

ApplySci:
It works on iphone ok, it's an Android-specific problem.

Based on debugging lines I added to the live `main` code [6f8157b](https://github.com/gdelataillade/alarm/commit/6f8157b6f4a72a58945da74440d9da8b524ffd1f) at [https://github.com/gdelataillade/alarm/blob/6f8157b6f4a72a58945da74440d9da8b524ffd1f/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt#L125](AlarmService.kt#L125):

[...]

PR #319, opened 2025-01-14
fix for #311

ApplySci:
The alarm listener doesn't fire if app is in foreground (Android only) 

The problem is a collision between Alarm and Firebase Cloud Messaging. When I retrieve my FCM token, it starts a new background isolate. Alarm, seeing this new engine, attaches to it, instead of the main engine. And this means that Alarm can no longer send messages to the app, so the listener fails. This fix adds a check to ensure that we only attach to the main engine, and not any subsequent isolates.


gdelataillade:
Hi @ApplySci 

Thanks for the PR ! I'll review it ASAP.

gdelataillade (approved):
LGTM

gdelataillade:
Added the fix to `5.0.0-dev.7`.

PR #325, opened 2025-01-27
Avoid creating duplicate timers and avoid scheduling alarms with the …

orkun1675:
…same ID


ios/Classes/api/AlarmApiImpl.swift
```diff
@@ -163,13 +170,12 @@ public class AlarmApiImpl: NSObject, AlarmApi {
...
-                let time = currentTime + dateTime.timeIntervalSinceNow
-                audioPlayer.play(atTime: time)
-            }
-
-            if let alarm = self.alarms[id], let delayInSeconds = alarm.triggerTime?.timeIntervalSinceNow {
+            if let alarm = self.alarms[id], let delayInSeconds = alarm.triggerTime?.timeIntervalSinceNow, !(alarm.timer?.isValid ?? false) {
+                NSLog("[SwiftAlarmPlugin] Rescheduling alarm with id \(id)")
+                
+                // We need to make sure the existing timer is invalidated to prevent duplicate task triggers.
+                alarm.timer?.invalidate()
```
orkun1675, line 177: This is the critical line in this PR, PTAL.

ios/Classes/api/AlarmApiImpl.swift
```diff
@@ -163,13 +170,12 @@ public class AlarmApiImpl: NSObject, AlarmApi {
 
         for id in ids {
             NSLog("[SwiftAlarmPlugin] Background check alarm with id \(id)")
-            if let audioPlayer = self.alarms[id]?.audioPlayer, let dateTime = self.alarms[id]?.triggerTime {
```
orkun1675, line 166: Is this manual alarm ring trigger needed? Note that it ignores volume settings so if need to do this for consistency I think we should find a way to call the `handleAlarmAfterDelay` method.

ios/Classes/services/AlarmStorage.swift
gdelataillade, line 1: Why are you removing the `AlarmStorage.swift`?

I created it when I implemented the **stop button in the notification**. This feature allows users to long-press the notification to show the stop button, which stops the alarm without requiring the user to open the app. Since the Flutter engine is not active in the background, this functionality must be implemented natively.

By the way, I recently tested this feature and discovered that the app opens when tapping on the stop button (which shouldn’t happen). I’ll investigate this issue soon.

orkun1675: I see. Why does iOS need to delete the alarm from storage when stopping it?

- It seems it will be stopped and removed anyway at next launch, by the Flutter code here: https://github.com/gdelataillade/alarm/blob/748679b4eca5ce2f389ad47f4374ae66da543bec/lib/alarm.dart#L64
- We should avoid cross-platform storage modification if we can: i.e. iOS shouldn't modify Flutter storage and vice versa

ios/Classes/api/AlarmApiImpl.swift
gdelataillade: I added this manual alarm ring trigger to increase the likelihood of the alarm ringing, especially for alarms scheduled for a long time in the future. However, I’m not entirely certain if it’s truly necessary.

ios/Classes/api/AlarmApiImpl.swift
gdelataillade: LGTM, but the only way to be sure is to run some long alarm tests. I'll keep you updated.

gdelataillade (commented):
Running some tests, I'll keep you posted.

gdelataillade:
### Tests:

- [x]  3 hours
- [ ] 12 hours
- [ ] 24 hours

PR #336, opened 2025-02-23
Android - Stop alarm on notification dismiss

ahamSel:
I noticed when notification is dismissed there isn't a way to stop the alarm unless through a patch in the app. This could be a big issue, so why not stop it when swiped away/dismissed the same way as the stop button


gdelataillade:
Thanks I'll review this tomorrow !

gdelataillade (approved):
LGTM

Issue #328, opened 2025-02-10
Notification Banners disappears after the alarm sound ended


taha-elkholy:
**Alarm plugin version**
*5.0.1 & 4.0.11* 

**Describe the bug**
*When setting an alarm it works as expected on both Android and ios but on ios only after the alarm sound ended the notification banner disappear when I swipe down, and no notification exists*

**To Reproduce**
1. *Set new Alarm*
2. *After the sound ends swipe down to see the notification banner*
3. *There are no notifications*

**Expected behavior**
*After the alarm sound ends; and when the user swipes down, a notification banner must be found*

**Device info**
*iPhone 15 Pro Max, iOS 17.5*

taha-elkholy:
in this file =>ios/Classes/api/AlarmApiImpl.swift 
just commenting this line 417 
`// NotificationManager.shared.dismissNotification(id: id)`
fix the issue with me

orkun1675:
Did you maybe set `loop = false`?

taha-elkholy:
```
final alarmSettings = AlarmSettings(
        id: notificationModel.id,
        dateTime: notificationModel.dateTime,
        notificationSettings: NotificationSettings(
          title: notificationModel.title,
          body: notificationModel.body,
          stopButton: S.current.stop,
          icon: 'logo_24',
        ),
        assetAudioPath: 'assets/sounds/${notificationModel.customSound}.m4a',
        loopAudio: false,
        vibrate: false,
        volumeSettings: VolumeSettings.fixed(
          volume: notificationModel.volume,
          volumeEnforced: true,
        ),
        androidFullScreenIntent: false,
      );
      await Alarm.set(
        alarmSettings: alarmSettings,
      );
```

orkun1675:
This is WAI. With the current codebase the alarm notification is dismissed when `loopAudio: false` and the alarm sound ends.

Can you describe semantically why you believe this behaviour should change?

taha-elkholy:
We use alarm to set alerts for prayer times, so the azan sound should be broadcast only once. When the azan sound is over, if the user does not hear the azan sound and stops the azan himself, in this case we need the notification to remain there to inform the user of the prayer time through it, otherwise the user will not know that he has entered the prayer time.

taha-elkholy:
Sure, I will do

tuanbs:
[...]

PR #339, opened 2025-03-01
Keep notification after alarm ends

taha-elkholy:
# Add iOS notification persistence option

## Description
Adds the ability to control whether notifications persist after an alarm ends on iOS devices. This feature allows users to maintain alarm notifications in the notification center even after the alarm sound ends.

Implements the feature requested in [#328](https://github.com/gdelataillade/alarm/issues/328)

## Changes
- Added `keepNotificationAfterAlarmEnds` property to `NotificationSettings` model
- Added corresponding field to `NotificationSettingsWire` in Pigeon API
- Generated updated platform bindings

## Example Usage
```dart
NotificationSettings(
  title: 'Wake Up',
  body: 'Time to start your day',
  keepNotificationAfterAlarmEnds: true, // Notification will remain in iOS notification center
);
```

## Technical Details
- iOS-specific feature (no effect on Android)
- Default value is `false` to maintain backward compatibility
- Property is serialized through platform channel communication

## Testing
- [x] Test notification persistence on iOS when `keepNotificationAfterAlarmEnds` is true
- [x] Verify no impact on Android devices
- [x] Verify backward compatibility with existing alarm configurations

## Documentation
Added comprehensive documentation with:
- Clear iOS-only feature indication
- Default value specification
- Behavior description for both true/false states

## Breaking Changes
None. This is a backward-compatible feature addition.

### DIFF
--- a/ios/Classes/models/NotificationSettings.swift
+++ b/ios/Classes/models/NotificationSettings.swift
@@ -4,6 +4,7 @@ struct NotificationSettings: Codable {
     var title: String
     var body: String
     var stopButton: String?
+    var keepNotificationAfterAlarmEnds: Bool = false
 
     static func from(wire: NotificationSettingsWire) -> NotificationSettings {
         // NotificationSettingsWire.icon is ignored since we can't modify the
@@ -11,7 +12,8 @@ struct NotificationSettings: Codable {
         return NotificationSettings(
             title: wire.title,
             body: wire.body,
-            stopButton: wire.stopButton
+            stopButton: wire.stopButton,
+            keepNotificationAfterAlarmEnds: wire.keepNotificationAfterAlarmEnds ?? false
         )
     }
 
@@ -19,7 +21,8 @@ struct NotificationSettings: Codable {
         return NotificationSettings(
             title: json["title"] as! String,
             body: json["body"] as! String,
-            stopButton: json["stopButton"] as? String
+            stopButton: json["stopButton"] as? String,
+            keepNotificationAfterAlarmEnds: json["keepNotificationAfterAlarmEnds"] as? Bool ?? false
         )
     }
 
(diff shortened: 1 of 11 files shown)


ios/Classes/models/NotificationSettings.swift
```diff
@@ -4,30 +4,34 @@ struct NotificationSettings: Codable {
...
 
     static func from(wire: NotificationSettingsWire) -> NotificationSettings {
         // NotificationSettingsWire.icon is ignored since we can't modify the
         // notification icon on iOS.
         return NotificationSettings(
             title: wire.title,
             body: wire.body,
-            stopButton: wire.stopButton
+            stopButton: wire.stopButton,
+            keepNotificationAfterAlarmEnds: wire.keepNotificationAfterAlarmEnds ?? false
```
orkun1675, line 16: We don't need to make the NotificationSettingsWire backwards compatible since instances of that object are short lived (created and disposed during native-flutter communication).

So `NotificationSettingsWire.keepNotificationAfterAlarmEnds` should not be nullable.

lib/model/notification_settings.dart
```diff
@@ -56,6 +57,18 @@ class NotificationSettings extends Equatable {
   /// Defaults to `null`.
   final String? icon;
 
+  /// Whether to keep the notification visible after the alarm sound ends.
+  ///
+  /// **iOS only feature - has no effect on Android.**
```
orkun1675, line 62: How does the notification behave on Android? Can we implement Android support as well? At a minimum we should describe in the docstring what the behaviour is.

pigeons/alarm_api.dart
orkun1675, line 1: Just wanted to make sure you saw line 3 of this file?

orkun1675 (changes requested):

pigeons/alarm_api.dart
taha-elkholy: Yse, I have do

ios/Classes/models/NotificationSettings.swift
taha-elkholy: Done, check this file again please

orkun1675:
@taha-elkholy any updates on this PR?

Note that we just merged a large refactor of the Swift side, you will need to rebase this PR.

taha-elkholy:
@orkun1675  I was finished all the above notes on 3 Mar; but for now after your last merge i have to fix the new conflicts

taha-elkholy:
this use case fixed in the alarm 5.1.3 so we do not need to modify it here

PR #350, opened 2025-03-24
Add option to set notification icon color

nicolausgr:
This PR adds the option to set `iconColor` within `NotificationSettings` so that the android notification icon gets the desired background color in expanded view of the notifications instead of the default blue that may seem out of place compared to the theme of many apps.

The `iconColor` can be null and defaults to null, thus nothing will break in existing codebases. But if it is passed a `Color`, then a string representation of this color's alpha, red, green, blue values are passed through `NotificationSettingsWire` to android where they are restored to an `Int` that represents the color and passed to `setColor` of the `notificationBuilder`.

 Since the `NotificationSettingsWire` is changed to contain the color values, 3 auto-generated files are changed by `dart run pigeon`.

I understand that you may want to implement this differently or not implement it at all. I was about to submit a feature request but I thought that it could be more helpful if I made the changes myself and submit a PR. So here it is.

(By the way, thanks for this package that seems to be the most credible flutter option for showing time-precise scheduled notifications on android!)


gdelataillade (approved):
LGTM

pigeons/alarm_api.dart
```diff
@@ -73,12 +73,20 @@ class NotificationSettingsWire {
...
+    required this.iconColorRed,
+    required this.iconColorGreen,
+    required this.iconColorBlue,
   });
 
   final String title;
   final String body;
   final String? stopButton;
   final String? icon;
+  final String? iconColorAlpha;
```
orkun1675, line 86: Why did we decide to represent colors components as strings?

nicolausgr: Fair point.
The initial intention was to pass the full color as an integer but it didn't work out between dart and kotlin. So then I used a string representation of the color and made it work and then, to make it simpler, I broke it to color components. You are right that it's a better decision to pass the components as doubles so please check #353.

PR #351, opened 2025-03-25
Fix self in concurrently-executing code

LesYampolskyi:
This PR fixes Swift Compiler Error (Xcode): 
Reference to captured var 'self' in concurrently-executing code

I don't know what the exact issue is, and maybe there is a better solution. Without this fix, the app will not build ios. 

[Similar issue](https://forums.swift.org/t/how-is-this-a-reference-to-a-var-in-concurrently-executing-code/66299/2) with discussions.


gdelataillade (approved):
LGTM

PR #352, opened 2025-03-26
Fix kotlin color after replacing flutter's Color deprecated properties

nicolausgr:
Flutter Color has deprecated properties `alpha`, `red`, `green`, `blue` and suggests using `a`, `r`, `g`, `b`.

However, `a`, `r`, `g`, `b` return floating-point values (between 0.0 and 1.0) while the deprecated properties returned integers between 0 and 255.

So, after you replaced the deprecated properties, the strings passed through `NotificationSettingsWire` represent floats and in `NotificationSettings.kt` you need to convert them to float now (and not integer).

Kotlin's Color argb() method can get 4 floats as it could also get 4 integers so no other change is needed.

I am sorry for not having replaced the deprecated properties in the initial PR, my fault.


gdelataillade (approved):
LGTM

gdelataillade:
Thanks for the quick fix !

PR #353, opened 2025-03-28
Icon color components as doubles through pigeon

nicolausgr:
As @orkun1675 pointed out, it's a better decision to pass the color components as doubles through pigeon.

In Kotlin, we still need to convert them from `Double` to `Float` for passing them to `Color.argb()` method.


orkun1675 (approved):
Thanks for the improvement!

Issue #357, opened 2025-04-21
Swiping away the application from recent apps stops the alarm.


sweakpl:
**Alarm plugin version**
5.1.2

**Describe the bug**
Swiping away the application from recent apps stops the alarm.

**To Reproduce**
1. Launch the example app.
2. Schedule an alarm.
3. When alarm rings, swipe away the app from recent applications - this will stop the alarm.

**Expected behavior**
The alarm should not stop as it has not been explicitly stopped by the user.
I would expect this behavior to be at least modifiable so that the alarm can continue ringing after swiping.

**Device info**
Android

orkun1675:
This is WAI so should be marked as a feature request IMO.

Wouldn't the user panic if they quit the app and the alarm keeps ringing?

sweakpl:
I've marked this as a bug because this behavior is not anything default in other alarm apps (and definitely not in my system alarm app) - swiping the app DOES NOT stop the alarm. There is always a foreground service running on Android and you can access the alarm via the notification.

I am not sure how exactly that works on iOS however.

Nevertheless I think that either:
a. this should be fixed so that alarm doesn't stop when the user swipes the app from recents,
b. or this behavior should be modifiable so that I can explicitly set that the alarm should be immune to swipes.

sweakpl:
Pretty sure this part should be within a condition guarded by a user specified parameter:

```kotlin
override fun onTaskRemoved(rootIntent: Intent?) {
    Log.d(TAG, "App closed, stopping alarm.")
    unsaveAlarm(alarmId)
    stopSelf()
    super.onTaskRemoved(rootIntent)
}
```

gdelataillade:
I agree with @orkun1675 — some users might panic if they try to stop the alarm by swiping away the app and the alarm keeps ringing, even though it can still be stopped via the notification.

To address this, we could introduce a new parameter: stopAlarmOnAppKill, which would default to true. This would allow developers to control whether the alarm should automatically stop when the app is killed.

Since on iOS the alarm naturally stops when the app is terminated (as the native code can no longer run), this would be an Android-only feature.

@sweakpl would you be interested in helping contribute to this feature?

sweakpl:
Sure, I will give it a go and come back with a PR if everything goes right 👍

gdelataillade:
Awesome !

sweakpl:
I've created a PR with this change: https://github.com/gdelataillade/alarm/pull/359

sweakpl:
[...]

PR #359, opened 2025-04-27
Added the androidStopAlarmOnTermination parameter

sweakpl:
The parameter `androidStopAlarmOnTermination` is used to specify whether to stop the alarm when an Android task is terminated by e.g. swiping away the app from the recent apps list.

This parameter has no effect on iOS.


orkun1675:
AFAIK there is no way to keep audio playing after the app is quit by the user (by swiping up from task manager) on iOS. Maybe you can check if iTunes or YouTube is even able to achieve this?

So this feature request might not be doable.

sweakpl:
Sorry, I can't really verify that.

However, as stated by @gdelataillade in https://github.com/gdelataillade/alarm/issues/357, this feature is supposed to be Android-only, so iOS should not be a concern.

sweakpl:
@orkun1675 I've added the disclaimer about the limitation on iOS to the README file.

gdelataillade (approved):
LGTM

sweakpl:
Great, thanks so much!

PR #363, opened 2025-05-22
Simplify Android plugin lifecycle

orkun1675:


orkun1675:
Yep, we've been testing this in prod for a while now, no reported issues, but would be great if you can test as well.

gdelataillade:
I encountered no issues during my tests.

PR #381, opened 2025-09-05
Fix Alarm.ringing stream not firing on Android 

### LINKED ISSUE #380: Alarm.ringing listener never fires when alarm rings (but Alarm.scheduled works)
After migrating from the deprecated `ringStream` to the new` Alarm.ringing` / `Alarm.scheduled` streams, the A`larm.ringing` listener never emits.

1. `Alarm.scheduled` does emit when I schedule/cancel alarms.
2. The alarm sound/notification triggers on time, but `Alarm.ringing` does not fire, so my app can’t navigate to the ringing screen.

This reproduces consistently in foreground, background, and after app restart.

NaoKreuzeder:
- Added `platform_timers.dart` to unify iOS and Android backup timer logic, reducing code duplication.
- Removed `ios_timers.dart` as it’s replaced by `platform_timers.dart`.
- Fixed issue where `Alarm.ringing` stream fails to emit on Android, ensuring reliable navigation to the ring screen.
- **Follow-up**: Refactored `ios_alarm.dart` and `android_alarm.dart` into `BaseAlarm` with subclasses to reduce duplication, per orkun1675's suggestion.
- Updated `platform_timers.dart` and `alarm.dart` to use instance methods for `IOSAlarm` and `AndroidAlarm`.
- Tested on Android emulator/device and iOS simulator with alarms in foreground and background.

Closes #380 (references https://github.com/gdelataillade/alarm/issues/380#issuecomment-3256844728)

### DIFF
--- a/lib/alarm.dart
+++ b/lib/alarm.dart
@@ -104,8 +104,8 @@ class Alarm {
     await AlarmStorage.saveAlarm(alarmSettings);
 
     final success = iOS
-        ? await IOSAlarm.setAlarm(alarmSettings)
-        : await AndroidAlarm.set(alarmSettings);
+        ? await IOSAlarm().setAlarm(alarmSettings)
+        : await AndroidAlarm().setAlarm(alarmSettings);
 
     if (success) {
       _scheduled.add(_scheduled.value.add(alarmSettings));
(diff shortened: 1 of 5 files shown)


orkun1675:
LGTM, thank you for the conribution @NaoKreuzeder ! 

Let's merge this asap.

As a follow up: notice how similar the implementation in `IOSAlarm` and `AndroidAlarm` classes are. Can we create a base class to avoid this much code repetition and make it easier to maintain and debug?

orkun1675:
By the way, you might need to run `dart format .` to satisfy the build checks.

NaoKreuzeder:
> As a follow up: notice how similar the implementation in `IOSAlarm` and `AndroidAlarm` classes are. Can we create a base class to avoid this much code repetition and make it easier to maintain and debug?

Good suggestion — the implementations are indeed quite similar. I’ll refactor them into a common base class to reduce duplication and improve maintainability.

orkun1675 (approved):

gdelataillade (approved):
LGTM

PR #396, opened 2026-05-19
fix(android): use STREAM_ALARM instead of STREAM_MUSIC for audio routing and volume enforcement

lw-manoj:
## Problem

On Android, alarm audio should use `STREAM_ALARM` (controlled by the alarm volume slider). Two files currently use `STREAM_MUSIC` (media volume):

**`AudioService.kt`**: `MediaPlayer` is created with no `AudioAttributes`, causing Android to default it to `STREAM_MUSIC`. Real-world effect: the media volume slider controls alarm audio; the alarm volume slider has no effect.

**`VolumeService.kt`**: All `setStreamVolume` calls reference `STREAM_MUSIC`, so volume enforcement targets the wrong stream. Additionally, `setVolume()` called `setStreamVolume` unconditionally — the `volumeEnforced` flag only gated the periodic re-enforcement loop, not the initial set. This causes the volume bar to appear and system volume to change even when `volumeEnforced: false`.

## Fix

- **`AudioService.kt`**: Set `AudioAttributes(USAGE_ALARM)` on the `MediaPlayer` (API 21+), with `setAudioStreamType(STREAM_ALARM)` fallback for API 19–20.
- **`VolumeService.kt`**: Add early return when `!volumeEnforced`. Replace all `STREAM_MUSIC` with `STREAM_ALARM`.

## Why `STREAM_ALARM` is correct

`AudioManager.STREAM_ALARM` has been available since API 1 — there is no version-compatibility reason to use `STREAM_MUSIC`. Android's own alarm clock apps use `STREAM_ALARM`. The `setAudioAttributes` API (API 21) is the only version-specific concern, handled by the fallback above.

## Notes

No existing unit tests cover `AudioService` or `VolumeService`. Adding instrumented tests for these would require an Android instrumented test setup, which is out of scope for this fix.

**Tested on:** Pixel 9a, Android 16

### DIFF
--- a/android/src/main/kotlin/com/gdelataillade/alarm/services/AudioService.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/services/AudioService.kt
@@ -1,8 +1,11 @@
 package com.gdelataillade.alarm.services
 
 import android.content.Context
+import android.media.AudioAttributes
+import android.media.AudioManager
 import android.media.MediaPlayer
 import android.media.RingtoneManager
+import android.os.Build
 import com.gdelataillade.alarm.models.VolumeFadeStep
 import java.util.concurrent.ConcurrentHashMap
 import java.util.Timer
@@ -84,6 +87,17 @@ class AudioService(private val context: Context) {
                     }
                 }
 
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                    setAudioAttributes(
+                        AudioAttributes.Builder()
+                            .setUsage(AudioAttributes.USAGE_ALARM)
+                            .setContentType(AudioAttributes.CONTENT_TYPE_SONIFICATION)
+                            .build()
+                    )
+                } else {
+                    @Suppress("DEPRECATION")
+                    setAudioStreamType(AudioManager.STREAM_ALARM)
+                }
                 prepare()
                 isLooping = loopAudio
                 start()
--- a/android/src/main/kotlin/com/gdelataillade/alarm/services/VolumeService.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/services/VolumeService.kt
@@ -23,11 +23,11 @@ class VolumeService(context: Context) {
     private var volumeCheckRunnable: Runnable? = null
 
     fun setVolume(volume: Double, volumeEnforced: Boolean, showSystemUI: Boolean) {
-        val maxVolume = audioManager.getStreamMaxVolume(AudioManager.STREAM_MUSIC)
-        previousVolume = audioManager.getStreamVolume(AudioManager.STREAM_MUSIC)
+        val maxVolume = audioManager.getStreamMaxVolume(AudioManager.STREAM_ALARM)
+        previousVolume = audioManager.getStreamVolume(AudioManager.STREAM_ALARM)
         targetVolume = (round(volume * maxVolume)).toInt()
         audioManager.setStreamVolume(
-            AudioManager.STREAM_MUSIC,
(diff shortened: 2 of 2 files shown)


gdelataillade:
@lw-manoj thanks for the contribution! I'll take a look at this as soon as possible.

gdelataillade (changes requested):
Thanks for the detailed write-up — the diagnosis of the `STREAM_MUSIC` vs `STREAM_ALARM` issue is spot-on, and parts 1 and 2 of this PR are clean wins. One concern on part 3.

### ✅ `AudioService.kt` — `AudioAttributes(USAGE_ALARM)`
LGTM. The API-21+ guard with `setAudioStreamType` fallback matches the existing pattern used for audio focus in `VolumeService.requestAudioFocus`.

### ✅ `VolumeService.kt` — `STREAM_MUSIC` → `STREAM_ALARM`
LGTM and overdue. Volume enforcement was operating on a different stream than the audio focus request, which has always been inconsistent.

### ⚠️ `VolumeService.kt` — `if (!volumeEnforced) return`
This part contradicts the documented contract of `VolumeSettings.volume` (`lib/model/volume_settings.dart:63-72`):

> Specifies the **system volume level** to be set when the alarm goes off. When the alarm is triggered, the system volume adjusts to this specified level. Upon stopping the alarm, the system volume reverts to its prior setting.

Current semantics: `volume` always sets the system volume; `volumeEnforced` only adds the periodic re-enforcement loop on top. With this early-return, `volume=0.7, volumeEnforced=false` becomes a no-op instead of "set the alarm stream to 70% for the duration of the alarm" — a behavior break for any caller relying on `volume` without enforcement.

The rationale in the PR description mentions the volume slider appearing, but that's already gated by the separate `showSystemUI` flag (the `FLAG_SHOW_UI` branch on line 32). Changing the system volume itself is the documented purpose of `volume`.

### Suggested change

Drop the `if (!volumeEnforced) return` and keep the original gating where only `startVolumeEnforcement` is conditional on `volumeEnforced`. The `STREAM_MUSIC` → `STREAM_ALARM` rename inside `setVolume` is independently correct and should stay. Roughly:

```kotlin
fun setVolume(volume: Double, volumeEnforced: Boolean, showSystemUI: Boolean) {
    val maxVolume = audioManager.getStreamMaxVolume(AudioManager.STREAM_ALARM)
    previousVolume = audioManager.getStreamVolume(AudioManager.STREAM_ALARM)
    targetVolume = (round(volume * maxVolume)).toInt()
    audioManager.setStreamVolume(
        AudioManager.STREAM_ALARM,
        targetVolume,
        if (showSystemUI) AudioManager.FLAG_SHOW_UI else 0
    )

    if (volumeEnforced) {
        startVolumeEnforcement(showSystemUI)
    }
}
```

With that adjustment, happy to merge.

lw-manoj:
Thanks @gdelataillade for catching that — you're absolutely right. The volumeEnforced flag was never meant to suppress the initial volume set, only the periodic re-enforcement loop. The showSystemUI/FLAG_SHOW_UI flag is already the correct knob for controlling volume bar visibility. I've updated the fix accordingly.

gdelataillade (approved):
Thanks for the quick turnaround — the updated  preserves the documented contract while still fixing the stream routing. LGTM.

gdelataillade (approved):
Thanks for the quick turnaround — the updated `VolumeService.setVolume` preserves the documented contract while still fixing the stream routing. LGTM.

gdelataillade:
Published to `5.3.0`, thanks for your contribution @lw-manoj !

PR #399, opened 2026-05-20
feat(android): expose showSystemUI through VolumeSettings so consumers can suppress the volume bar

### LINKED ISSUE #398: bug(android): showSystemUI flag is hardcoded to true in AlarmService — consumers cannot suppress the volume bar
## Summary

`VolumeService.setVolume()` correctly accepts a `showSystemUI` parameter that controls whether the system volume bar appears when the alarm sets the volume (`FLAG_SHOW_UI`). However, this parameter is never reachable by consumers because `AlarmService.kt` hardcodes it to `true`:

```kotlin
// AlarmService.kt:40
private var showSystemUI: Boolean = true
```
It is never read from AlarmSettings or VolumeSettings. Every call to volumeService?.setVolume(...), volumeService?.restorePreviousVolume(...) passes this hardcoded value — making the showSystemUI parameter on VolumeService effectively dead.

## Impact
Consumers have no way to prevent the system volume bar from appearing when an alarm triggers and adjusts the volume. For many alarm UX patterns (silent background volume enforcement, waking the user without extra UI noise), this is undesirable.

## Root Cause
showSystemUI is absent from the entire settings chain:

* VolumeSettings (Dart model) — no field
* VolumeSettingsWire (pigeon definition) — no field
* VolumeSettings.kt (Android model) — no field
* AlarmService.kt — hardcoded private var showSystemUI: Boolean = true

## Expected Behaviour
Consumers should be able to pass showSystemUI: false via VolumeSettings to suppress the volume bar when the alarm sets/restores the system volume.

## Suggested Fix
* Add showSystemUI (default true for backward compatibility) to VolumeSettingsWire in pigeons/alarm_api.dart
* Add it to the Dart VolumeSettings model and all named constructors (fixed, fade, staircaseFade)
* Regenerate pigeon bindings: dart run pigeon --input pigeons/alarm_api.dart && dart format .
* Add showSystemUI to VolumeSettings.kt on Android
* Replace the hardcoded value in AlarmService.kt with alarmSettings.volumeSettings.showSystemUI

lw-manoj:
## Problem

Discovered during review of #396. `AlarmService.kt` hardcodes `showSystemUI = true`,
making it impossible for consumers to prevent the system volume bar from appearing
when the alarm sets or restores volume:

```kotlin
// AlarmService.kt
private var showSystemUI: Boolean = true
```
`VolumeService.setVolume()` already accepted showSystemUI correctly and gated
`FLAG_SHOW_UI` on it — the parameter was simply never reachable from the public API.

## Fix
Threads `showSystemUI` (default `true` for backward compatibility) through the full stack:

* `pigeons/alarm_api.dart` — added to `VolumeSettingsWire`
* `lib/model/volume_settings.dart` — added to `VolumeSettings` and all named constructors (`fixed`, `fade`, `staircaseFade`)
* Regenerated platform bindings (Dart + Kotlin + Swift)
* `VolumeSettings.kt` — added field, read from wire
* `AlarmService.kt` — removed hardcoded `true`, reads from `alarmSettings.volumeSettings.showSystemUI` at alarm start

## Usage

```kotlin
VolumeSettings.fixed(
  volume: 0.8,
  showSystemUI: false, // suppress the volume bar
)
```

Fixes #398.
Tested on: Pixel 9a, Android 16

### DIFF
--- a/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
@@ -176,6 +176,7 @@ class AlarmService : Service() {
 
         // Retrieve whether the alarm should be stopped on task termination
         shouldStopAlarmOnTermination = alarmSettings.androidStopAlarmOnTermination
+        showSystemUI = alarmSettings.volumeSettings.showSystemUI
 
         // Acquire a wake lock to wake up the device
         val wakeLock = (getSystemService(Context.POWER_SERVICE) as PowerManager)
--- a/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
@@ -136,6 +136,9 @@ class AlarmService : Service() {
             }
         }
 
+        // Read showSystemUI before any volume calls that depend on it
+        showSystemUI = alarmSettings.volumeSettings.showSystemUI
+
         // Set the volume if specified
         if (alarmSettings.volumeSettings.volume != null) {
             volumeService?.setVolume(
@@ -176,7 +179,6 @@ class AlarmService : Service() {
 
         // Retrieve whether the alarm should be stopped on task termination
         shouldStopAlarmOnTermination = alarmSettings.androidStopAlarmOnTermination
-        showSystemUI = alarmSettings.volumeSettings.showSystemUI
 
         // Acquire a wake lock to wake up the device
         val wakeLock = (getSystemService(Context.POWER_SERVICE) as PowerManager)
--- a/android/src/main/kotlin/com/gdelataillade/alarm/generated/FlutterBindings.g.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/generated/FlutterBindings.g.kt
@@ -171,7 +171,8 @@ data class VolumeSettingsWire (
   val volume: Double? = null,
   val fadeDurationMillis: Long? = null,
   val fadeSteps: List<VolumeFadeStepWire>,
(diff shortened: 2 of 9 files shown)


gdelataillade (commented):
Heads-up on a migration risk before merge.

`VolumeSettings.kt` is `@Serializable`, and the new `showSystemUI: Boolean` field has no default on the Kotlin data class:

```kotlin
data class VolumeSettings(
    ...
    val volumeEnforced: Boolean,
    val showSystemUI: Boolean   // no default
)
```

In `AlarmSettings.kt` (around line 75), persisted alarms are decoded via:

```kotlin
Json.decodeFromJsonElement(VolumeSettings.serializer(), it)
```

Existing on-device alarms scheduled by 5.3.0 and earlier have a `volumeSettings` JSON object without `showSystemUI`, so kotlinx-serialization will throw `MissingFieldException` on upgrade and those alarms will fail to ring. The Dart side handles this correctly (`?? true` in `fromJson`, defaults in all constructors), so the public API is non-breaking — but Android persistence isn't.

One-line fix:

```kotlin
val showSystemUI: Boolean = true
```

With that default, the PR is fully backward-compatible. Nice catch on the underlying bug otherwise — clean threading through all layers.

lw-manoj:
Thank you @gdelataillade for catching this — you're absolutely right. The `@Serializable` data class without a default would throw `MissingFieldException` on upgrade for any alarm persisted before this change. The legacy path fix in `AlarmSettings.kt` alone wasn't sufficient. Added `= true` as the default directly on the field so `kotlinx-serialization` handles it gracefully across all deserialization paths.

gdelataillade (approved):
LGTM thanks

gdelataillade:
Added to version `5.4.0` thanks!

lw-manoj:
Thank you!

PR #401, opened 2026-05-29
feat(android): add preferConnectedAudioDevice flag to route audio to …

### LINKED ISSUE #400: Alarm sound ONLY plays through the speaker, bypassing earphones (5.4.0)
**Alarm plugin version**
5.4.0 (Worked fine in 5.2.1)

**Describe the bug**
After updating the package from 5.2.1 to 5.4.0, the alarm sound plays only through the device's external speaker, bypassing connected earphones entirely. Absolutely no sound comes through the earphones (wired or Bluetooth). In the previous version (5.2.1), the sound was correctly routed to the earphones when they were connected.

**Expected behavior**
When earphones are connected, the alarm sound should play through the earphones. If forcing the external speaker is an intended feature for wake-up alarms, could you please add an option (e.g., choosing the audio stream type like STREAM_MUSIC) so developers can route the sound to earphones?

**Device info**
Samsung Galaxy S24+ (SM-S926N), Android 15 (SDK 35)

lw-manoj:
…earphones or Bluetooth

STREAM_ALARM / USAGE_ALARM (introduced in 5.4.0) forces audio to the built-in speaker by Android design, bypassing connected earphones and Bluetooth devices. This is correct for wake-up alarms but is a regression for developers who need external device routing.

Adds preferConnectedAudioDevice (default false) to AlarmSettings. When true, both AudioService and VolumeService switch to STREAM_MUSIC + USAGE_MEDIA, routing to a connected earphone or Bluetooth device if present and falling back to the speaker if not. The trade-off is that the media volume slider controls the volume instead of the alarm slider.

Also fixes androidStopAlarmOnTermination being absent from AlarmSettings copyWith() and props — two callers relying on equality or copyWith for this field would have seen silent incorrect behaviur.

Threads the flag through the full stack: AlarmSettingsWire pigeon definition → Dart AlarmSettings model → regenerated platform bindings → Android AlarmSettings.kt (with = false default for safe deserialization of persisted alarms) → AlarmService.kt → AudioService.kt and VolumeService.kt.

Fixes #400.

Tested on: Pixel 9a, Android 16


gdelataillade:
Thanks @lw-manoj, the fix looks great — clean threading through every layer, backward compat is solid, and nice catch on the stale `alarm_settings.g.dart` (it was silently dropping `androidStopAlarmOnTermination` from persisted alarms).

Two small things before I merge:

### 1. Audio focus stream consistency

`VolumeService.requestAudioFocus()` still hardcodes `USAGE_ALARM` / `STREAM_ALARM` even when `preferConnectedAudioDevice` is `true`. The MediaPlayer playback uses `USAGE_MEDIA` but the focus request stays alarm-typed, which is inconsistent and may affect how other apps duck/yield. Could you thread the flag into `requestAudioFocus()` as well (and through the `AlarmService` call site)?

### 2. README docs

Could you add a row for `preferConnectedAudioDevice` to the AlarmSettings table in `README.md` (around line 89, next to `androidStopAlarmOnTermination`)? Something like:

```
| preferConnectedAudioDevice | `bool` | If true, routes alarm audio to a connected earphone or Bluetooth device when present, falling back to the speaker. Uses the media volume slider instead of the alarm slider. Android-only. Disabled by default. |
```

Once those are in I'll merge. Thanks again for the fast turnaround!

lw-manoj:
Thanks @gdelataillade for reviewing and catching the missing update. I will fix it later tonight and update the PR.

lw-manoj:
Thanks for the thorough review @gdelataillade. Both points addressed:

1. **Audio focus consistency**:  `requestAudioFocus()` now branches on `preferConnectedAudioDevice` the same way `AudioService` does: `USAGE_MEDIA` (API 26+) and `STREAM_MUSIC` (pre-26) when `true`, `USAGE_ALARM` / `STREAM_ALARM` when `false`. Audio focus and playback are now consistent.

2. **README**: Added `preferConnectedAudioDevice` row to the `AlarmSettings` table, with wording adjusted to "Has no effect on iOS" rather than "Android-only" to be more precise and consistent with the existing doc style.

gdelataillade:
Thank you! Added to version `5.4.1` !

Issue #403, opened 2026-06-05
Add support for scheduling multiple alarms at the same second


Fi2zz:
**Is your feature request related to a problem? Please describe.**  
`Alarm.set` currently replaces an existing alarm when another alarm is scheduled for the same `dateTime` down to the second, even if the two alarms have different `id` values.

This behavior is documented, but it is too restrictive for some valid use cases. For example, an app may need to schedule multiple reminders at the exact same time for different tasks, entities, or grouped events. In those cases, only the last scheduled alarm survives, which makes the scheduling behavior less flexible than the `id` model suggests.

**Describe the solution you'd like**  
Please add a new `AlarmSettings` option to allow alarms with different `id` values to coexist even when they share the same `dateTime` down to the second.

For example:

```dart
AlarmSettings(
  id: 1,
  dateTime: dateTime,
  allowAlarmOverlap: true,
  allowSameSecondScheduling: true,
)
```

Expected behavior:
- alarms with the same `id` should still replace each other
- the current behavior should remain the default for backward compatibility
- when `allowSameSecondScheduling` is enabled, alarms with different `id` values should not replace each other just because they are scheduled for the same second

This would fit well with the current model-level configuration style, such as `allowAlarmOverlap`.

**Describe alternatives you've considered**  
The current workaround is to artificially offset one alarm by at least one second, but that changes the intended trigger time and is not suitable for apps that need precise simultaneous alarms.

Another possible API design would be adding a parameter on `Alarm.set(...)`, but an `AlarmSettings` field seems more consistent with the existing package design and easier to persist/serialize across platforms.

**Additional context**  
The current same-second replacement happens in the shared Dart layer before native scheduling is reached, so this appears to be a package-level policy rather than a hard native limitation.

Relevant locations:
- `lib/alarm.dart`
- `lib/utils/extensions.dart`

From investigation:
- Android native scheduling is primarily keyed by `id`
- iOS native scheduling also primarily replaces by `id`
- this request is only about scheduling coexistence
- it is separate from `allowAlarmOverlap`, which controls whether multiple alarms may ring at the same time after they are triggered

A local repro project was created and confirms the current behavior: when two alarms with different `id` values are scheduled for the same second, only the last one remains scheduled.

### Implementation proposal

Add a new field on `AlarmSettings`:

```dart
final bool allowSameSecondScheduling;
```

Recommended behavior:
- same `id` -> always replace existing alarm
- different `id` + same second + `allowSameSecondScheduling == false`
[...]

PR #404, opened 2026-06-10
feat: add allowSameSecondScheduling for same-second alarm coexistence

Fi2zz:
- Dart: add allowSameSecondScheduling to AlarmSettings
- iOS: implement ringing queue with LIFO order, currentAlarmId protection
- Android: implement ringing queue with LIFO order, per-id audio completion listener
- Example: add toggles and demo buttons for both strategies, keep stopped alarms in list

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,25 @@ You can also listen to the `Alarm.updateStream` to know when an alarm is added, 
 
 To avoid unexpected behaviors, if you set an alarm for the same time, down to the second, as an existing one, the new alarm will replace the existing one.
 
+If you need to schedule multiple alarms with different ids for the exact same second, you can set `allowSameSecondScheduling` to `true`:
+
+```dart
+AlarmSettings(
+  id: 1,
+  dateTime: dateTime,
+  allowSameSecondScheduling: true,
+  // ...
+)
+```
+
+When `allowSameSecondScheduling` is enabled:
+- Alarms with the **same id** still replace each other
+- Alarms with **different ids** can be scheduled for the same second
+- How they ring depends on `allowAlarmOverlap`:
+  - `allowAlarmOverlap = false` (default): Alarms ring **sequentially** one after another — just like the iOS system Clock app. The first alarm rings first; when it stops, the next queued alarm starts ringing automatically.
+  - `allowAlarmOverlap = true`: Alarms ring **concurrently** — the later alarm will override the previous one and continue ringing.
+- These two options are independent and can be combined as needed
+
 ## 📱 Example app
 
 Don't hesitate to check out the [example's code](https://github.com/gdelataillade/alarm/tree/main/example), and take a look at the app:
--- a/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt
@@ -39,6 +39,8 @@ class AlarmService : Service() {
     private var alarmStorage: AlarmStorage? = null
     private var showSystemUI: Boolean = true
     private var shouldStopAlarmOnTermination: Boolean = true
+    private val ringingQueue = mutableListOf<Int>()
+    private val queuedAlarmSettings = mutableMapOf<Int, AlarmSettings>()
 
     override fun onCreate() {
         super.onCreate()
@@ -92,6 +94,36 @@ class AlarmService : Service() {
             return START_NOT_STICKY
         }
 
+        // If another alarm is already ringing
+        if (ringingAlarmIds.isNotEmpty()) {
+            if (!alarmSettings.allowAlarmOverlap) {
+                // Queue for sequential ringing (like iOS system Clock app)
+                ringingQueue.add(id)
+                queuedAlarmSettings[id] = alarmSettings
+                Log.d(TAG, "Alarm $id queued because another alarm is already ringing.")
+                return START_NOT_STICKY
+            }
+            // allowAlarmOverlap == true: continue to ring, later alarm overrides previous one
+        }
+
+        ringAlarm(id, alarmSettings)
+        return START_STICKY
+    }
+
+    private fun ringAlarm(id: Int, alarmSettings: AlarmSettings) {
+        alarmId = id
+
+        // Build the notification
+        val notificationHandler = NotificationHandler(this)
+        val appIntent =
+            applicationContext.packageManager.getLaunchIntentForPackage(applicationContext.packageName)
+        val pendingIntent = PendingIntent.getActivity(
+            this,
+            id,
+            appIntent,
+            PendingIntent.FLAG_IMMUTABLE or PendingIntent.FLAG_UPDATE_CURRENT
+        )
+
         val notification = notificationHandler.buildNotification(
             alarmSettings.notificationSettings,
             alarmSettings.androidFullScreenIntent,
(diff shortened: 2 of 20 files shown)


gdelataillade (changes requested):
Hey @Fi2zz, thanks a lot for the PR — I took a deep look. The wire/model plumbing is complete and correct on all three layers, and the iOS side genuinely addresses the concurrency concerns from #403: the `currentAlarmId` guard in `AlarmRingManager.stop(id:)` even fixes an existing race on main (stopping an overridden alarm used to kill the newer alarm's ring). The per-id audio completion listeners on Android are also exactly what I asked for.

That said, I found a few blocking issues — mostly on the Android side — that need to be addressed before this can merge.

## Blocking issues

### 1. Android: the ringing queue dies on the most common stop path

The queue lives in `AlarmService`, but stopping a **ringing** alarm from Dart (`Alarm.stop(id)` — i.e. the stop button in the app's own ring screen) goes through `AlarmApiImpl.stopAlarm()`, which calls `context.stopService(...)`. That destroys the whole service, and your `onDestroy()` clears `ringingQueue` / `queuedAlarmSettings`.

Concretely, in your own "3 seq" demo: alarm 1 rings, alarms 2 and 3 are queued. If the user stops alarm 1 from the app's ring screen (not the notification button), the service is destroyed and alarms 2 and 3 are silently dropped — and they stay in storage as ghost "scheduled" alarms that will never ring (their PendingIntents already fired), with no `alarmStopped` notification to Dart. Sequential ringing currently only works when alarms are stopped via the notification's Stop button (`handleStopAlarmCommand` → `stopAlarm`, which has the dequeue logic).

Suggested fix: route `AlarmApiImpl.stopAlarm` through the service's `stopAlarm` (e.g. deliver the existing `STOP_ALARM` action to the running service instead of calling `stopService`), so the queue survives and promotion runs.

### 2. Android: a dismissed queued alarm can still ring

If the user stops a **queued** alarm from Dart (it's not in `ringingAlarmIds`, so the `stopService` branch is skipped), it gets unsaved from storage but remains in the service's queue. When the currently ringing alarm is stopped via the notification button, the deleted alarm is promoted and rings anyway.

You handled this correctly on iOS (`guard self.alarms[nextId] != nil`), but Android has no equivalent check. Suggested fix: validate against `AlarmStorage` before promoting, and/or remove the id from `ringingQueue` / `queuedAlarmSettings` when a queued alarm is stopped.

### 3. Sequential queueing is not gated behind the new flag

Both queue checks (Android `AlarmService.onStartCommand` and iOS `AlarmManager.ringAlarm`) only test `allowAlarmOverlap` — not `allowSameSecondScheduling`. This changes behavior for **every existing user with default settings**: today, an alarm that fires while another is ringing is dropped and Dart is notified it stopped; with this PR it's silently queued and rings whenever the first one stops — possibly much later (an 8:05 alarm firing mid-8:00-alarm now rings at 8:20 when the user finally stops the first one).

The issue explicitly stated "the current behavior should remain the default for backward compatibility", and the README addition documents sequential ringing as part of the new flag, which isn't what the code does. Please gate the queueing behind the flag — or if you think queue-when-busy deserves to be its own opt-in (it's arguably a separate feature from same-second scheduling), give it its own flag. Either way it must not change default behavior.

### 4. Unrelated regression in `checkAlarm` (lib/alarm.dart)

The PR removes the branch that re-emits an actively-ringing alarm to the `ringing` stream on app restart, replacing it with an unconditional `stop()`. On Android this path is load-bearing: with `androidStopAlarmOnTermination: false`, the foreground service keeps ringing after the app is killed; when the user reopens the app, main re-surfaces the ring UI — with this change, the alarm is silenced the moment the app launches. This change isn't mentioned in the PR description and isn't needed for the feature. Please revert it.

## Minor

- CI will fail: `flutter analyze` exits 1 on this branch — 4 info-level lints, all in the example app (line length ×2, redundant argument value, missing `const`).
- LIFO ordering: with three queued alarms they ring in reverse arrival order on both platforms. FIFO would match user intuition better — or at least document the LIFO choice.
- The example-app rework (`alarmHistory` keeping stopped alarms visible, `Navigator.removeRoute`) is fine for demoing, though a bit more opinionated than the feature needs.

The iOS half is close to merge-ready — the remaining work is concentrated on the Android stop paths and the gating. Happy to re-review once these are addressed!

gdelataillade (commented):
Thanks for the quick turnaround @Fi2zz — I re-reviewed `9e19fd8` and it's almost there. Verified resolved:

- ✅ **Gating**: queueing now requires `allowSameSecondScheduling == true` on both platforms; with default settings the behavior is identical to main (drop + notify). The README is accurate now too.
- ✅ **Dismissed queued alarms**: `stopAlarm` removes the id from the queue, and `triggerNextQueuedAlarm()` validates against `AlarmStorage` before promoting — mirrors the iOS guard nicely.
- ✅ **`checkAlarm` revert**: exact restore of the ring-recovery path.
- ✅ **Lints**: `flutter analyze` exits 0 on the branch now.

One remaining issue with the `AlarmApiImpl.stopAlarm` fix:

### `context.startService()` is now called unconditionally — background crash risk

The queue does survive Dart-initiated stops now (👍), but `startService(stopIntent)` runs on **every** `Alarm.stop()`, including when no alarm is ringing and the service isn't running. If `Alarm.stop()` is called while the app is backgrounded with no foreground service active — e.g. canceling a scheduled alarm from an FCM/background isolate handler — `startService` throws `IllegalStateException: Not allowed to start service Intent` on API 26+. The old code never hit this because it only called `stopService`, and only when ringing.

Side effects of the same change: stopping a future alarm spins up the whole service just to deliver `STOP_ALARM` (×N for `stopAll`), and non-ringing stops send `alarmStopped` to Flutter twice (once from `AlarmApiImpl`, once from the service's `unsaveAlarm`) — harmless but noisy.

**Suggested fix** — do what `AlarmReceiver` already does for the notification stop button: call the running service directly and skip the intent entirely:

```kotlin
// Deliver the stop to the running service so it can clean up ringing
// alarms and dequeue if needed. If the service isn't running, there is
// nothing to ring/dequeue and storage cleanup below is sufficient.
AlarmService.instance?.handleStopAlarmCommand(id)
```

No background-start restriction, no service churn, and `handleStopAlarmCommand` → `unsaveAlarm` → `stopAlarm` runs the same dequeue/promotion logic you added.

That's the only blocker left — everything else looks merge-ready. 🙏

Fi2zz:
@gdelataillade Got it, thanks for the feedback. The issue has been fixed now.

gdelataillade:
Heads-up @Fi2zz: a second review pass surfaced two more Android issues introduced by the queue/listener changes, both confirmed against the current head. Since they were small and this PR is otherwise ready, I pushed `ddf2dcd` directly to your branch (thanks for leaving maintainer edits enabled) rather than asking for another round:

1. **Non-looping alarms lost their completion cleanup.** `playAudio()` internally calls `stopAudio(id)` to release any previous player for that id — and since the per-id listener change, `stopAudio()` also removes `onAudioCompleteListeners[id]`. Because `AlarmService.ringAlarm` registers the listener *before* calling `playAudio`, the listener was wiped immediately, so non-looping alarms never stopped vibration / restored volume / abandoned audio focus when the audio ended. Fixed by splitting the player teardown into a private `releaseMediaPlayer(id)` that `playAudio` uses internally, while the public `stopAudio(id)` keeps removing the listener. Also cleared the listener map in `cleanUp()` to match the other maps.

2. **Queued alarms overwrote the "current alarm" id used by `onTaskRemoved`.** `onStartCommand` assigned `alarmId = id` before the queue decision, so if alarm A was ringing and B got queued, swiping the app away cleaned up B instead of A — wrong `alarmStopped` event, and the ringing alarm left as a ghost in storage. Since `ringAlarm()` already sets `alarmId`, the early assignment is simply removed.

Kotlin compiles cleanly and `flutter analyze` is unaffected. With these two on top of your fixes, I'm happy with the state of the PR.

gdelataillade:
Will be added in the next release!

Issue #392, opened 2026-01-16
AGP 9 Support


ihadabs:

Adityapanther:
@gdelataillade could you please check this pull #406  ?

PR #405, opened 2026-06-26
Add support for AGP9

jmietzITZ:
+ bump compileSdk to 35 (a dependency already requires it)
+ add support for AGP9 with and without builtin kotlin support

Only tested so far inside my flutter app. It would be great if others could also test this code in their apps to be sure there are no issues!

Relevant Issue: #392

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,3 +17,25 @@ jobs:
       - run: flutter --version
       - run: flutter pub get
       - run: flutter analyze
+
+  build-android:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # JDK 21 (not 17) so compileKotlin runs on a JDK newer than the Java target,
+      # matching Android Studio's bundled JBR and catching JVM-target mismatches.
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '21'
+      - uses: subosito/flutter-action@v2
+        with:
+          channel: 'stable'
+          cache: true
+      - uses: gradle/actions/setup-gradle@v4
+      - run: flutter --version
+      - name: Build example APK
+        working-directory: example
+        run: |
+          flutter pub get
+          flutter build apk --debug
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 5.6.0
-* [Android] Update to AGP9 and built-in Kotlin.
+* [Android] Added support for Android Gradle Plugin 9 and its built-in Kotlin toolchain, while keeping compatibility with AGP 8. The Kotlin Gradle plugin is now applied conditionally, following Flutter's official plugin migration guide.
+* [Android] Bumped `compileSdk` from 34 to 35 and the build's Kotlin version from 2.1.0 to 2.2.10.
+* [Android] Minimum Kotlin Gradle plugin version for consuming apps is now 2.0.0 (already required by Flutter 3.44+). No API or runtime behavior changes.
+* Updated the example app to AGP 9.0.1 / Gradle 9.1.0.
 
 ## 5.5.0
 * [iOS] Added Swift Package Manager support.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.6.0
+* [Android] Update to AGP9 and built-in Kotlin.
+
 ## 5.5.0
 * [iOS] Added Swift Package Manager support.
 * Added `AlarmSettings.allowSameSecondScheduling` to allow multiple alarms to ring in the same second.
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -26,31 +26,32 @@ apply plugin: 'com.android.library'
 
 def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
 def builtInKotlin = providers.gradleProperty('android.builtInKotlin').map { it.toBoolean() }.orElse(true).get()
-if (agpMajor < 9) {
+def shouldApplyKotlinAndroidPlugin = agpMajor < 9 || !builtInKotlin
+
+// AGP 9 compiles Kotlin via its built-in Kotlin support unless the app opts out.
+if (shouldApplyKotlinAndroidPlugin) {
     apply plugin: 'kotlin-android'
-} else if (!builtInKotlin) {
-    apply plugin: 'com.android.built-in-kotlin'
 }
 apply plugin: 'kotlinx-serialization'
 
 android {
-    namespace 'com.gdelataillade.alarm.alarm'
+    namespace = 'com.gdelataillade.alarm.alarm'
 
-    compileSdkVersion 35
+    compileSdk = 35
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
     }
 
-    if (agpMajor < 9) {
+    if (shouldApplyKotlinAndroidPlugin) {
         kotlinOptions {
             jvmTarget = '17'
         }
     }
 
     defaultConfig {
-        minSdkVersion 19
+        minSdk = 19
     }
 
     buildTypes {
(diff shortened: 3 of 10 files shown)


gdelataillade (changes requested):
Thanks for tackling this 🙏 — AGP 9 support is overdue and the conditional approach here is the right shape. I built and tested the branch against **AGP 9.0.1 / Gradle 9.1 / JDK 17** in both Kotlin configurations before reviewing. Summary: the **published plugin (`android/build.gradle`) is solid and non-regressing**, but the PR can't merge as-is because the **example app no longer builds** under current Flutter.

### What I verified ✅
- **`:alarm` compiles on AGP 9 with `builtInKotlin=false`** (the `kotlin-android` branch) — `BUILD SUCCESSFUL`.
- **`kotlinx-serialization` works under built-in Kotlin without `kotlin-android`** — I reproduced the exact gradle setup in an isolated AGP 9 library; `.serializer()` and `Json.encodeToString` resolve and compile. This was my biggest worry (it would otherwise have been a silent runtime `SerializationException`), so good to have it confirmed.
- The `agpMajor < 9 || !builtInKotlin` logic and the `.orElse(true)` default correctly match AGP 9's behavior (built-in Kotlin on by default, bundled KGP 2.2.10). Removing the `src/main/kotlin` sourceSet is safe.

### 🔴 Blocker: the example app fails to build
`flutter build apk` on this branch fails with:
```
Inconsistent JVM-target compatibility detected for tasks
'compileDebugJavaWithJavac' (17) and 'compileDebugKotlin' (21).
```
Flutter's migrator forces `android.builtInKotlin=false`, so removing `kotlinOptions { jvmTarget = '17' }` from `example/android/app/build.gradle` leaves Kotlin defaulting to JVM 21 against Java 17. Re-adding it fixes the build (verified — `✓ Built app-debug.apk`):
```gradle
kotlinOptions {
    jvmTarget = '17'
}
```
Worth mirroring the plugin's conditional guard so it's correct whether or not built-in Kotlin is active.

### ⚠️ Heads-up: migration is partial (not a blocker)
Because the plugin still applies `kotlinx-serialization`, Flutter continues to report `alarm` under *"plugins that apply Kotlin Gradle Plugin (KGP) … future versions of Flutter will fail to build."* There's no KGP-free way to use serialization today, so this is expected — but it means the full built-in-Kotlin path isn't actually reachable end-to-end via current Flutter yet (forcing `builtInKotlin=true` NPEs in Flutter's own gradle plugin, and sibling plugins like `flutter_fgbg`/`permission_handler` aren't migrated). Every current consumer hits the `builtInKotlin=false` branch, which works. Just flagging so expectations are clear.

### Minor / optional
- `android/build.gradle` still pins `classpath 'com.android.tools.build:gradle:8.7.0'` — works only because the app's AGP shadows it; a plugin shouldn't pin AGP.
- Example still pins `ndkVersion "27.0.12077973"` (AGP 9 plugins now want `28.2.13676358`).
- `help/INSTALL-ANDROID.md` still references `compileSdkVersion 34` and applying `org.jetbrains.kotlin.android` — outdated for AGP 9.
- compileSdk 34 → 35 is fine, but worth a CHANGELOG note since consumers on 34 may get a warning.

Could you re-add the example app's `jvmTarget`? Once the example builds I'm happy to merge. 👍

Adityapanther:
I am facing this issue, when would you fix it ?

jmietzITZ:
thanks for the feedback, will incorporate the changes probably later today

Adityapanther:
any update ?

gdelataillade:
Thanks for the update! I re-tested [1629cef](https://github.com/gdelataillade/alarm/commit/1629cefad6880879939a06eec1b53bae57b20e98) against AGP 9.0.1 / Gradle 9.1 / JDK 17. The original blocker isn't resolved yet, and this commit introduces a new CI failure — flagging both before this goes further.

### 🔴 Blocker still open: example app doesn't build
This commit doesn't touch `example/android/app/build.gradle`, so `kotlinOptions { jvmTarget = '17' }` is still missing. Rebuilding gives the same failure as before:
```
Inconsistent JVM-target compatibility detected for tasks
'compileDebugJavaWithJavac' (17) and 'compileDebugKotlin' (21).  →  BUILD FAILED
```
It's actually now locked in more explicitly: `example/android/gradle.properties` was updated to commit `android.builtInKotlin=false` / `android.newDsl=false` — which is the exact configuration that hits this failure. Re-adding the `kotlinOptions` block (as suggested in my first review) still fixes it.

### 🔴 New: CI (`flutter analyze`) will fail
The `very_good_analysis: ^7.0.0 → ^10.3.0` bump in `pubspec.yaml` introduces two new `discarded_futures` lints in files this PR doesn't otherwise touch:
- `lib/service/alarm_storage.dart:114`
- `lib/src/platform_timers.dart:113`

Verified directly:
- `main` (very_good_analysis 7.0.0): `flutter analyze` → *No issues found!* (exit 0)
- this branch (very_good_analysis 10.3.0): `flutter analyze` → 2 issues (exit 1)

Since `flutter analyze` is the only job in `.github/workflows/main.yml`, this PR will fail CI as-is.

### 🟠 Worth double-checking: `pubspec.yaml` SDK constraint change
```diff
- sdk: ">=3.0.0 <4.0.0"
+ sdk: ^3.8.0
```
This raises the **minimum Dart SDK for every consuming app** to 3.8.0 — looks like incidental fallout from a `pub`/template refresh rather than something needed for AGP 9 support. If intentional it should get a CHANGELOG callout (it's a breaking change for anyone below 3.8), otherwise worth reverting to keep this PR scoped to the AGP 9 migration. Same question for the `json_annotation: ^4.9.0 → ^4.12.0` bump (a shipped dependency) and the `pigeon 25→27` dev bump.

### ✅ Addressed since last review
- CHANGELOG entry and 5.6.0 version bump — thanks for adding these.

To get this mergeable: re-add the example's `jvmTarget`, and either revert `very_good_analysis` to `^7.0.0` or fix the two `discarded_futures` call sites so `flutter analyze` is clean again. Happy to take another pass as soon as those land.

jmietzITZ:
ran into some issues, there is some stuff to clean up for AGP9 :D
The example app probably still needs to turn builtin kotlin support off, as some other dependencies dont completely support it yet.
Also upgrading the kotlin and some dependency versions introduced some breaking changes which I'm working on fixing.

gdelataillade:
Thanks for getting this to the finish line @jmietzITZ 🙏 — the conditional approach here is exactly right. To wrap it up I pushed three commits to the branch (`maintainerCanModify` made that easy — hope that's alright):

1. **Fixed the example build.** Re-added `kotlinOptions { jvmTarget = '17' }`, guarded by the same AGP/built-in-Kotlin conditional as the plugin, so it's correct whether or not built-in Kotlin is active. Also switched `ndkVersion` to `flutter.ndkVersion` so it tracks the SDK. This clears the `Inconsistent JVM-target compatibility (17 vs 21)` blocker.
2. **Reverted the unrelated bumps** so the PR stays scoped to AGP 9: the `sdk`/`flutter` constraints, `json_annotation`, `pigeon` (generated code is still v25.5.0, so no regen needed), and `very_good_analysis` (10.x added two `discarded_futures` lints that broke `flutter analyze`). Version stays 5.6.0.
3. **Added a `build-android` CI job** (the example was never built in CI before) and refreshed the docs/changelog.

### Validation
Built and compiled locally on Flutter 3.44.4 / JDK 17:

| Scenario | Result |
|---|---|
| Example APK — AGP 9.0.1 / Gradle 9.1 (built-in Kotlin) | ✅ builds |
| `:alarm` — AGP 8.9.1 / Gradle 8.11.1 / KGP **2.1.0** | ✅ compiles |
| `:alarm` — AGP 8.9.1 / KGP **2.0.0** (floor) | ✅ compiles |
| `flutter analyze` / `flutter pub publish --dry-run` | ✅ clean |

The AGP 8 legs confirm the biggest risk of the Kotlin bump (`2.1.0 → 2.2.10`): the `kotlinx-serialization` compiler subplugin still compiles the plugin's `@Serializable` models against apps on KGP 2.0.0/2.1.0 — no metadata-version errors. Documented minimum KGP is now 2.0.0 (Flutter 3.44+ already enforces this).

### One expected non-issue
Flutter still lists `alarm` under *"plugins that apply Kotlin Gradle Plugin (KGP)…"* — its detector is a regex on the (conditional) `apply plugin` line. There's no KGP-free way to use `kotlinx-serialization` today, and every current consumer takes the working `builtInKotlin=false` path, so this is cosmetic and unchanged from released versions. Flagging so nobody files it as a regression.

Once CI is green I'll squash-merge. Thanks again!

PR #406, opened 2026-07-08
AGP9 and built in kotlin support added

Adityapanther:
1) #392 
2) #405


gdelataillade:
Closing in favor of #405, which landed AGP 9 support using Flutter's official *conditional* built-in-Kotlin pattern — it keeps AGP 8 working and doesn't raise the minimum Kotlin/Flutter versions for existing apps.

Thank you for this PR @Adityapanther 🙏 — it was genuinely useful for comparing approaches, and your example-app modernization (removing `kotlin-android`, the AGP 9 DSL) matched what we ended up shipping. The main reasons we went with #405 instead:

- It keeps the runtime dependencies unchanged. Bumping `kotlinx-serialization-json` to 1.11.0 pulls in kotlin-stdlib 2.3.x, which fails to compile for consumers on Kotlin < 2.2 with an "incompatible metadata version" error.
- It applies the Kotlin plugin conditionally (`agpMajor < 9`), so apps still on AGP 8 / Flutter 3.41–3.43 keep building; the unconditional removal here requires Flutter ≥ 3.44.

Appreciate you pushing on this — it helped get AGP 9 support over the line. 🙌

Issue #395, opened 2026-03-26
Add Swift Package Manager support


abdallahshaban557:
Hello,

Flutter is migrating to [Swift Package Manager](https://www.swift.org/documentation/package-manager/). Please [add Swift Package Manager support](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin) to your plugin.

SwiftPM will be the default for all Flutter apps in the next stable release (3.44). To ensure your plugin remains compatible and prevents build failures for your users, please migrate to SwiftPM

Flutter will eventually deprecate and then remove support for CocoaPods. Adding Swift Package Manager support future-proofs your plugins.

### **How to add Swift Package Manager support to your plugins**

You can find the migration guide here: [https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin)

If you run into issues or have questions, please reach out to the Flutter team. You can send a message to the #hackers-ios channel on [Flutter's discord](https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md) or [open a GitHub issue](https://github.com/flutter/flutter/issues/new/choose).

Thank you for your wonderful contributions to the Flutter ecosystem!

### **Why is Flutter migrating to Swift Package Manager?**

[CocoaPods is now in maintenance mode](https://blog.cocoapods.org/CocoaPods-Support-Plans/).

Swift Package Manager support is one of Flutter's most requested features: [flutter#33850](https://github.com/flutter/flutter/issues/33850).

Flutter's Swift Package Manager integration has several benefits:

1. Access to the Swift package ecosystem. Flutter plugins can use the growing ecosystem of [Swift packages](https://swiftpackageindex.com/).  
2. Simplifies Flutter installation. Swift Package Manager is bundled with Xcode. In the future, you won't need to install Ruby and CocoaPods to target iOS or macOS.

gdelataillade:
You’re right that Swift Package Manager support is an important enhancement for the plugin, and I do plan to add it. I can’t prioritize this immediately, but it’s on the roadmap for a future update, while keeping CocoaPods compatibility during the transition.

[...]

PR #407, opened 2026-07-12
#395: path issue fixed for swiftPM 

Adityapanther:
1) #395 
2) path issue fixed for swiftPM and ios example update for swiftPM


Adityapanther (commented):
LGTM

Issue #211, opened 2024-06-06
Action buttons (dismiss, snooze)


ShadichyDev:
**Is your feature request related to a problem? Please describe.**
There's no dismiss button and I can't swipe the notification away

**Describe the solution you'd like**
Add snoozeActionLabel and dismissActionLabel at least

**Describe alternatives you've considered**
I have to press power button to stop ringing /vibrating

**Additional context**
Android 12, api level 31

![Screenshot_20240604-164749](https://github.com/gdelataillade/alarm/assets/74492447/18bbac93-188b-4b9d-88e1-8f1ad41bb047)

gdelataillade:
Hi @ShadichyDev,

The alarm plugin uses Android’s Foreground Service to ensure the alarm can trigger at any time, even if the app is killed. However, this requires making the notification non-dismissible for Android 12+ due to Android’s rules.

Regarding notification action buttons (stop & snooze), they are on my to-do list. However, as I mostly work alone on the plugin and have limited time, it may take a while to implement.

I am closing this issue as it is a duplicate of issue #58. I will keep you updated there.

ShadichyDev:
I did not get the term "non-dismissible". Isn't it just invoke 'stopAlarm' when presssing Dismiss button if present?

gdelataillade:
When I said "non-dismissible" I meant that you can't swipe away the notification. But yes, when I'll add notification action buttons, it will call `stopAlarm`. But it's not that simple because I will have to do it native-side, that's where I manage the notifications. I also have to update the local storage. Let me know if you have any question.

Issue #409, opened 2026-07-24
Publish 5.6.0: AGP 9 / built-in Kotlin fix is on main but not on pub.dev


gonzapersonalsite:
Hi! First of all, thank you for this package — the Android side does the
genuinely hard part (foreground service, AlarmManager, audio focus), and
having it maintained as MIT saved us from writing and owning all of that
ourselves. Much appreciated.

This is not a bug report: the issue below is **already fixed on `main`**.
I'm just flagging that the published artifact still carries it, in case
that's useful for prioritising a release.

## Context

The AGP 9 / built-in Kotlin work landed on `main` in 63e1ce9
("feat(android): AGP 9 and built-in Kotlin support", 2026-07-11), but the
newest version on pub.dev is still 5.5.0, so consumers don't get it yet.

## What consumers on 5.5.0 currently hit

**1. Flutter's KGP warning on every build**

```
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): alarm
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.
```

Worth noting this is a true positive for the published artifact rather
than the known false-positive detector bug — `alarm-5.5.0/android/build.gradle`
applies the plugin unconditionally:

```groovy
apply plugin: 'com.android.library'
apply plugin: 'kotlin-android'      // no agpMajor / builtInKotlin guard
apply plugin: 'kotlinx-serialization'
```

`main` already guards this correctly.

**2. Build failure from `compileSdkVersion 34`**

5.5.0 pins `compileSdkVersion 34`, while its own transitive dependency
requires 35 or later, so a clean build fails:

```
Dependency ':flutter_fgbg' requires libraries and applications that
depend on it to compile against version 35 or later of the Android APIs.
:alarm is currently compiled against android-34.
```

Our workaround is forcing plugin modules up to the app's `compileSdk`
from the root `build.gradle.kts`. `main` already sets `compileSdk = 35`.

## Question

Is a 5.6.0 release on the roadmap? No rush at all — I noticed dab6c26
mentions built-in Kotlin still having end-to-end compatibility issues
with the Flutter tooling, so I completely understand if you're waiting
for the ecosystem to settle before publishing.

Happy to test a pre-release against a real device suite if that would
help. We run alarm delivery through integration tests on a physical
Android device, so I can report back on anything specific.

## Environment

- `alarm` 5.5.0 (resolved from pub.dev)
[...]

PR #411, opened 2026-07-25
feat(android): add a snooze action to the alarm notification

gonzapersonalsite:
### Problem

The notification carries a stop action and nothing else, so an alarm can only be deferred from inside the app — which defeats the point of a snooze, since the user has to unlock the device and navigate to it.

In #211 you wrote that action buttons were on your to-do list, and that the hard parts were doing it native-side where the notifications are managed, and updating local storage. That is what this does.

### Change

An alarm given `androidSnoozeDuration`, and a notification given a `snoozeButton` label, offers a snooze that stops the current ring and re-registers the alarm that far ahead through the ordinary scheduling path.

```dart
AlarmSettings(
  // ...
  androidSnoozeDuration: const Duration(minutes: 5),
  notificationSettings: const NotificationSettings(
    title: 'Wake up',
    body: '',
    stopButton: 'Stop',
    snoozeButton: 'Snooze',
  ),
);
```

Both are required: a label without a duration describes an action the platform cannot perform.

### Reported as a deferral, not a stop

A snooze produces `alarmSnoozed`, never `alarmStopped`. A snoozed alarm is still owed, and reporting it as a stop would tell the app the user dismissed something they asked to be reminded of again.

```dart
Alarm.snoozed.listen((snooze) {
  print('Alarm ${snooze.id} rings again at ${snooze.nextRingAt}');
});
```

### Why it is also stored

The button is normally pressed with **no Flutter engine running** — the notification is native and the process may not be up — so `alarmSnoozed` reaches nobody. Each deferral is therefore recorded in `AlarmStorage` and drained on the next `Alarm.init()` through `takeUnreportedSnoozes`.

Without that, an app tracking its own alarm state finds an alarm whose instant moved with nothing explaining why, and a reconciliation pass would move it back — silently undoing the snooze.

Receiving both the live report and the drained record applies the same deferral twice, which changes nothing.

### Compatibility

Alarms with no snooze duration behave exactly as before. The new `AlarmApi.takeUnreportedSnoozes` returns an empty list on iOS, where snoozing does not apply.

### A note on the generated files

`lib/model/*.g.dart` are edited by hand: `build_runner` on my toolchain emits null-aware elements your `sdk: ">=3.0.0 <4.0.0"` lower bound rejects, and fails on `Color? iconColor`. Please regenerate them with yours rather than taking mine. The pigeon outputs **are** properly regenerated from `pigeons/alarm_api.dart`.

### Testing

Example app builds and runs on Android; `flutter analyze` clean under your `very_good_analysis` config. Exercised in a real application on a Samsung SM-S928B (Android 16), including the no-engine path: snooze from the lock screen, kill the app, reopen, and the deferral is replayed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 5.7.0
+* [Android] Added a snooze action to the alarm notification, via `AlarmSettings.androidSnoozeDuration` and `NotificationSettings.androidSnoozeButton`. Snoozing defers the alarm rather than stopping it, and is reported on the new `Alarm.snoozed` stream as well as through `Alarm.scheduled`. A snooze taken with no Flutter engine running is applied on the next `Alarm.init()`.
 * [Android] Alarms can now be presented on the app's own activity by declaring an intent filter for `com.gdelataillade.alarm.action.RING`, instead of always opening the launcher activity.
+* Fixed `build_runner` being unable to regenerate `lib/model/*.g.dart`, which also removes the only `dart format` violation in the package.
 
 ## 5.6.0
 * [Android] Added support for Android Gradle Plugin 9, while keeping AGP 8 compatibility.
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ await Alarm.set(alarmSettings: alarmSettings)
 | androidStopAlarmOnTermination                       | `bool`                 | Whether to stop the alarm when an Android task is terminated. Enabled by default.                                                                                                                    |
 | preferConnectedAudioDevice                          | `bool`                 | If true, routes alarm audio to a connected earphone or Bluetooth device when present, falling back to the built-in speaker if not. Uses the media volume slider instead of the alarm slider. Has no effect on iOS. Disabled by default. |
 | payload                                             | `String?`              | Optional data sent with the alarm. Caller handles serialization and parsing.                                                                                                                         |
+| androidSnoozeDuration                               | `Duration?`            | How long the snooze action defers the alarm. Android only. With a `NotificationSettings.snoozeButton` label, the notification offers a snooze that stops the ring and re-registers the alarm this far ahead. No snooze if null. |
 | [notificationSettings](#notificationsettings-model) | `NotificationSettings` | Settings for notification title, body, icon, icon color and action buttons (only stop at the moment).                                                                                                |
 | [volumeSettings](#volumesettings-model)             | `VolumeSettings`       | Settings for alarm volume and fade durations.                                                                                                                                                        |
 
@@ -107,6 +108,7 @@ The property `androidStopAlarmOnTermination` works only on Android as on iOS the
 | title                          | `String`  | Title of the alarm notification.                                                   |
 | body                           | `String`  | Body of the alarm notification.                                                    |
 | stopButton                     | `String?` | Text shown in the stop button of the alarm notification. Button not shown if null. |
+| snoozeButton                   | `String?` | Text shown in the snooze button of the alarm notification. Android only. Shown only when `AlarmSettings.androidSnoozeDuration` is also set. |
 | icon                           | `String?` | Icon to display on the notification. Only customizable on Android.                 |
 | iconColor                      | `Color?`  | Color of the notification icon. Only customizable on Android.                      |
 | keepNotificationAfterAlarmEnds | `bool`    | Keeps the notification visible after the alarm sound ends. iOS only.               |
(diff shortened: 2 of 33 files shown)


gdelataillade:
Thanks for this, and sorry for the long silence. I've pushed directly to your branch rather than sending you round another review loop, since there was more to change than a comment would carry — full detail is in the commit messages.

**The one real bug.** Your reasoning for the durable record was exactly right: without it an app finds an alarm whose instant moved with nothing explaining why, and a reconciliation pass moves it back. What the record was missing was the last step — `_alarmSnoozed` emitted the streams but never wrote the shifted time to Dart's own store. So `checkAlarm()` still saw the original past time, found the alarm not ringing, and stopped it. The snooze cancelled itself on the next `Alarm.init()`. That's now the first of two named regression tests.

**What else changed, and why:**

- The snooze went through `AlarmService.instance?`, so with the service killed and its notification lingering it was a silent no-op. There's now a `SnoozeCoordinator` that needs no service, and `AlarmReceiver` always goes through it.
- Scheduling moved into an `AlarmScheduler` primitive that deliberately excludes the stop path. Your fresh `AlarmApiImpl` was actually the safe choice — its empty `alarmIds` is what stopped `setAlarm`'s dedupe branch from reporting the snooze to Flutter as a *stop*. Making that explicit rather than incidental.
- Ordering is inverted: the replacement is armed *before* the ring is silenced, so a failed reschedule leaves the alarm audibly ringing instead of silently cancelled.
- Seconds became milliseconds on the wire, and the destructive drain became `getPendingSnoozes` + `acknowledgeSnooze`, so a crash between reading and applying no longer loses the deferral.
- There's a one-minute minimum. Below about five seconds Android stops using `AlarmManager` and falls back to an in-process timer that survives neither termination nor cancellation.
- `snoozeButton` → `androidSnoozeButton`, since its sibling `stopButton` genuinely is cross-platform.
- #410 landed meanwhile, so the ring intent now also carries `alarmSnoozeLabel` — an app-owned ring screen can offer snooze too.

**Your `.g.dart` note was accurate**, and it turned out to be a repo bug rather than your toolchain: `build_runner` failed here too, and `--delete-conflicting-outputs` deleted the three model files. Fixed separately on `main` (pinned `json_serializable`, added a `JsonConverter` for `iconColor`), so codegen works again and `dart format` passes across the package for the first time.

The shape is still yours — `alarmSnoozed` rather than `alarmStopped`, the durable record, both fields required. CI is green and there are 66 tests, but none of the native ordering has run on real hardware.

Which is where your offer in #409 would genuinely help, if you're still up for it: snooze from the lock screen with the app killed, reopen before the deadline, and confirm it still rings. That's the exact path the regression test covers in Dart and can't cover natively. No rush.

---

**Update.** A second review pass turned up five more issues, all of them real. Fixed in `1668c39`; `pubspec.yaml` is now 5.7.0.

- **An already-applied marker left the alarm invisible.** If the process died after Dart persisted the new time but before the host acknowledged, the marker survived with storage already agreeing with it — so the idempotent early return also skipped putting the alarm back into `Alarm.scheduled` and re-arming its fallback timer in the fresh isolate. It still rang, but was missing from the stream that is supposed to list it. Idempotence now covers the write and the events only; process-local state is rebuilt every time, since it never survives a restart regardless of what storage says.
- **`copyWith` could not clear either new field**, so an editor UI had no way to remove a snooze once set. Both now take the same nullable callback as `payload`.
- **Snoozing did not restore the kill warning.** `ringAlarm` turns it off when the ringing alarm is the only one, and the snooze path re-armed the alarm without re-evaluating that — so killing the app mid-snooze produced no warning even though the alarm was pending again. The invariant now lives in a `WarningNotificationState` helper that every scheduling path goes through, with the text persisted rather than held in memory: the context that has to show it usually is not the one that was told it.
- **A duration-only alarm could still be deferred** by broadcasting `ACTION_SNOOZE` directly, even though it offers snooze nowhere in the UI and the README claimed the broadcast would be ignored. `AlarmReceiver` is exported, so that gate was the only thing enforcing the contract; it now requires the label too.
- **The example could not actually receive replayed snoozes.** `main()` awaits `Alarm.init()` before `runApp`, and `Alarm.snoozed` does not buffer, so the replay was emitted before the listener existed — my note in the example said otherwise and was simply wrong. `main()` now subscribes before `init()`, and the README documents the ordering, since it affects every consumer and not just the example.

69 tests, CI green on all three jobs.

The device-testing ask above still stands, and matters a little more now: these changes touched the scheduling path and the snooze gate.

gdelataillade:
Adding this to 5.7.0

PR #415, opened 2026-08-02
fix(android): don't ask to dismiss a secure keyguard when an alarm rings

### LINKED ISSUE #412: Android: requestDismissKeyguard forces authentication before the alarm UI can be used
## Summary

On a device with a secure lock screen (fingerprint, PIN or pattern), the user
has to authenticate before they can press Stop on the alarm screen. The alarm
rings and the screen turns on, but the system asks for a fingerprint first.

This happens even with `android:showWhenLocked="true"` and
`android:turnScreenOn="true"` set on the activity, as documented in Step 3 of
`help/INSTALL-ANDROID.md`.

## Steps to reproduce

1. Set up a secure lock screen (fingerprint or PIN).
2. Follow `INSTALL-ANDROID.md`, including `showWhenLocked` and `turnScreenOn`.
3. Schedule an alarm with `androidFullScreenIntent: true`.
4. Lock the screen and wait for the alarm.

## Expected

The alarm screen shows over the lock screen and its Stop / Snooze buttons work
without authentication, while the rest of the device stays locked. This is how
the AOSP and Google Clock alarms behave.

## Actual

The alarm screen is shown, but the system authentication prompt appears on top
of it. The user has to unlock the device before reaching the Stop button, which
is awkward at the exact moment an alarm app should be simplest to use.

## Cause

`AlarmPlugin.kt`, in `notificationObserver`:

```kotlin
if (it) {
    Log.d(TAG, "Making app visible on lock screen...")
    activity.setShowWhenLocked(true)
    activity.setTurnScreenOn(true)
    val keyguardManager =
        activity.applicationContext.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
    keyguardManager.requestDismissKeyguard(activity, null)
}
```

The first two calls display the activity over the keyguard, which is the goal.
The third one asks the system to dismiss the keyguard entirely, and the
[Android documentation](https://developer.android.com/reference/android/app/KeyguardManager#requestDismissKeyguard(android.app.Activity,%20android.app.KeyguardManager.KeyguardDismissCallback))
states that on a secure keyguard the user will be prompted to authenticate.

So `setShowWhenLocked(true)` already covers the "show over the lock screen"
case, and `requestDismissKeyguard` turns it into a stronger "unlock the whole
device" request, which triggers the prompt.

## Suggested fix

Make the keyguard dismissal opt-in, for example through an `AlarmSettings`
field defaulting to `false`:

```kotlin
if (it) {
    activity.setShowWhenLocked(true)
    activity.setTurnScreenOn(true)
    if (dismissKeyguard) {
        val keyguardManager =
            activity.applicationContext.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
        keyguardManager.requestDismissKeyguard(activity, null)
    }
}
```

A `false` default matches what users expect from an alarm app, and apps that
rely on the current behaviour can opt back in.

## Environment

- alarm 5.5.0
- Flutter 3.44.4
- Samsung Galaxy S23 (SM-S911U1), Android 16 / One UI

## Workaround

I currently keep a local copy of the package with that single line removed, and
it behaves as expected: the alarm screen is usable over the lock screen without
authentication, and leaving it still requires unlocking the device.

Happy to open a PR if the approach above sounds right to you.

samyak1409:
Closes #412.

## The problem

When an alarm starts, `AlarmPlugin.notificationObserver` does three things:

```kotlin
activity.setShowWhenLocked(true)
activity.setTurnScreenOn(true)
keyguardManager.requestDismissKeyguard(activity, null)
```

The first two are what the log line right above them describes — "Making app visible on lock screen" — and between them they put the alarm on top of the keyguard and wake the display. The platform's own docs for `turnScreenOn` are explicit that the flag turns the screen on by itself when the activity will be visible and resumed, which `showWhenLocked` guarantees:

> Specifies whether the screen should be turned on when the Activity is resumed. Normally an activity will be transitioned to the stopped state if it is started while the screen if off, but with this flag set the activity will cause the screen to turn on if the activity will be visible and resumed due to the screen coming on. […] This flag is normally used in conjunction with the showWhenLocked flag to make sure the activity is visible after the screen is turned on when the lockscreen is up.

The third call asks for something stronger — unlock the whole device — and on a secure keyguard the system [prompts the user to authenticate](https://developer.android.com/reference/android/app/KeyguardManager#requestDismissKeyguard(android.app.Activity,%20android.app.KeyguardManager.KeyguardDismissCallback)). So the alarm screen is up, but an authentication prompt sits on top of it, and Stop can't be reached until you unlock — at the exact moment an alarm app should be simplest to use. AOSP and Google Clock don't behave this way.

## The change

`requestDismissKeyguard` is now called only when the device isn't secured:

```kotlin
if (!keyguardManager.isDeviceSecure) {
    keyguardManager.requestDismissKeyguard(activity, null)
}
```

- **Secured device (PIN / pattern / password):** no dismissal request, so no prompt. The alarm shows over the lock screen and Stop works, while the rest of the device stays locked.
- **Unsecured device:** unchanged — the keyguard is dismissed as before, which costs the user nothing and leaves them on the alarm instead of the lock screen.

`isDeviceSecure` rather than `isKeyguardSecure` because the latter also reports true for a locked SIM, which has nothing to do with whether dismissing will prompt. It needs API 23; the observer already returns early below API 27 for `setShowWhenLocked`, so it's inside an existing version gate.

## On the flag suggested in the issue

#412 proposed an `AlarmSettings` field defaulting to `false`. I went with a straight fix instead, for two reasons — happy to switch if you'd rather have the knob:

1. A per-alarm setting can't reach here without a wider change. `notificationObserver` receives a bare `Boolean` from `AlarmRingingLiveData`, so it has no idea which alarm is ringing; plumbing a per-alarm flag in means changing what that LiveData carries, which touches the ring path in `AlarmService`.
2. With a `false` default the flag would exist only to restore the behaviour the issue reports as a bug — and this way apps that genuinely want the device unlocked can still call `requestDismissKeyguard` themselves from their own activity.

## Verification

- `flutter build apk` on `example/` — compiles, and `isDeviceSecure` is present in the built dex.
- `flutter analyze` — clean.
- `flutter test` — 48 passing (unchanged; the plugin has no Kotlin test harness, and this path needs a real Activity and KeyguardManager).

I haven't been able to verify the fixed path on hardware. The behaviour it fixes is reproducible: Pixel 6a / Android 17 and, per #412, Samsung S23 / Android 16 — two different manufacturers and OS versions.


gdelataillade:
Thanks! Will be added to `5.7.1`.

PR #417, opened 2026-08-02
feat(android): attach the alarm extras to the launcher fallback intent too

### LINKED ISSUE #416: [Android] The launcher fallback intent carries no alarm extras, so apps can't tell the alarm opened them
## Summary

When an alarm rings with `androidFullScreenIntent: true`, the app is launched — but on the default path it has no way to know the alarm is why. `ringPendingIntent` attaches the alarm extras only to a resolved `…action.RING` activity; the `getLaunchIntentForPackage` fallback gets none:

```kotlin
val ringIntent = resolveRingIntent()?.apply {
    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
    putExtra(EXTRA_ALARM_ID, id)
    putExtra(EXTRA_ALARM_TITLE, alarmSettings.notificationSettings.title)
    putExtra(EXTRA_ALARM_BODY, alarmSettings.notificationSettings.body)
    putExtra(EXTRA_ALARM_STOP_LABEL, alarmSettings.notificationSettings.stopButton)
} ?: applicationContext.packageManager
    .getLaunchIntentForPackage(applicationContext.packageName)
    ?: Intent()
```

That fallback intent is byte-identical to tapping the launcher icon — same action, same category, no extras — so from inside the app "the alarm opened us" and "the user opened us" are indistinguishable.

## Why it matters

It isn't only about which alarm is ringing (`Alarm.ringing` answers that). It's about **who opened the app**, which the ring stream can't answer, and apps need it to decide what to do when the ring ends.

Concretely: if the full-screen intent opened the app, most alarm apps want to get back out of the way once the alarm is stopped, rather than leave themselves sitting on top of whatever the user was doing — you unlock at 06:05 and your alarm app is the foreground app. But if the user had opened the app themselves a moment earlier, closing it would be wrong.

With no signal, the only thing left is a heuristic: watch the lifecycle state and treat "resumed very recently" as the intent's doing. The activity resume and the ring event race with no fixed order, so the window has to be guessed, and the guess is wrong in both directions. An extra removes the guess entirely.

## Suggested fix

Put the extras on whichever intent is used, not just the resolved RING one — the fallback branch is the one most apps take, since declaring a RING activity is opt-in:

```kotlin
private fun ringPendingIntent(id: Int, alarmSettings: AlarmSettings): PendingIntent {
    val ringIntent = (resolveRingIntent()
        ?: applicationContext.packageManager
            .getLaunchIntentForPackage(applicationContext.packageName)
        ?: Intent()).apply {
        addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
        putExtra(EXTRA_ALARM_ID, id)
        // …and the rest, as today
    }
    …
}
```

The constants are already public and already documented for the RING path, so this is the same contract reaching one more entry point. `PendingIntent` equality ignores extras, so nothing about matching or `FLAG_UPDATE_CURRENT` changes, and an app that ignores the extras is unaffected.

An app can then read it in `MainActivity.onCreate` / `onNewIntent` and pass it to Dart. If you'd rather expose it as a Dart-level signal instead of an intent extra, that works just as well for this use case — the extra is simply the smallest change that fits what's already there.

I'm happy to open the PR.

## Environment

- alarm 5.5.0 (checked against `main`, where the RING path was added)
- Flutter 3.44.8
- Pixel 6a, Android 17 (API 37)

samyak1409:
Closes #416.

## The problem

`ringPendingIntent` attaches the alarm extras only to a resolved `…action.RING` activity. Apps that don't declare one — which is most, since it's opt-in — get `getLaunchIntentForPackage`, and that intent is byte-identical to tapping the app icon: same action, same category, no extras. From inside the app, an alarm launch and a manual one are indistinguishable.

## Why that's worth fixing

Not for *which* alarm is ringing — `Alarm.ringing` answers that. For **who opened the app**, which it can't.

An alarm app generally wants to get back out of the way once the alarm is stopped, instead of sitting on top of whatever the user was doing: you unlock at 06:05 and your alarm app is the foreground app. But if the user had opened the app themselves a moment earlier, closing it would be wrong. Deciding that needs to know whether the full-screen intent is what put the app there.

With no signal the only option is a heuristic — watch the lifecycle state and treat "resumed very recently" as the intent's doing. The activity resume races the ring event with no fixed order, so the window is a guess, and it's wrong in both directions. One extra removes the guess.

## The change

The extras now go on whichever intent is used, rather than only the RING one:

```kotlin
val ringIntent = (resolveRingIntent()
    ?: applicationContext.packageManager
        .getLaunchIntentForPackage(applicationContext.packageName)
    ?: Intent()).apply {
    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
    putExtra(EXTRA_ALARM_ID, id)
    // …and the rest, unchanged
}
```

An app reads `alarmId` in `onCreate` / `onNewIntent` and knows.

**Nothing changes for anyone who ignores them.** The constants are already public and already documented for the RING path, so this is the same contract reaching one more entry point. `PendingIntent` equality ignores extras, so matching and `FLAG_UPDATE_CURRENT` behave exactly as before, and the RING branch is untouched — it just no longer has the extras written twice.

`FLAG_ACTIVITY_NEW_TASK` is now also applied to the fallback. `getLaunchIntentForPackage` already sets it, so that's a no-op rather than a behaviour change; it moved because the two branches now share one `apply`.

## Also in the diff

- README: a note under the extras table that they arrive on the launcher fallback too, so an app that doesn't want a separate alarm screen can still tell the two apart.
- CHANGELOG under 5.7.0, beside the RING entry this builds on.

## Verification

- `flutter build apk` on `example/` — compiles, and `alarmStopLabel` is present in the built dex.
- `flutter analyze` — clean.
- `flutter test` — 48 passing (unchanged; this path has no Kotlin test harness).

Not device-verified end to end. If you'd rather expose this as a Dart-level signal than an intent extra, that solves the same problem — the extra is just the smallest change that fits what's already there, and I'm happy to redo it either way.


samyak1409:
One note on the diff: I deliberately left out a CHANGELOG entry for this one. It would land on the same line of the unreleased 5.7.0 section as #415's, and with three open PRs touching a one-line section something has to give — this is the most speculative of the three, so it seemed the right one to keep friction-free. With it omitted, #414, #415 and this branch all merge into main cleanly in any order (checked locally). Happy to add the entry whenever you'd like it.

gdelataillade:
Thanks for this, and for #416 — the diagnosis is exactly right. The launcher fallback does get no extras, and `Alarm.ringing` genuinely can't tell you *who* opened the app.

I'm closing this one, because the capability already exists and I'd rather not put the extras on the launcher intent. `resolveRingIntent()` queries the whole package for any activity declaring the `RING` filter — it doesn't have to be a dedicated activity. Declaring it on your existing `MainActivity`, alongside the launcher filter, gets you the extras today with no plugin change:

```xml
<intent-filter>
    <action android:name="com.gdelataillade.alarm.action.RING" />
    <category android:name="android.intent.category.DEFAULT" />
</intent-filter>
```

That's a better signal than this PR produces: a launcher tap arrives as `ACTION_MAIN` and an alarm as `ACTION_RING`, so `intent.action` alone tells them apart — whereas with extras on both paths you'd be comparing extras instead.

I verified it end to end on a Pixel 8a / Android 16 before writing this. An alarm delivers `ACTION_RING` with `alarmId` (warm via `onNewIntent`, cold via `onCreate`), a launcher tap delivers `ACTION_MAIN` with no extras, and tapping the icon after an alarm-created task resumes the same activity rather than duplicating it.

The genuinely undocumented part was that the `RING` activity needn't be separate — that's on me. I've written it up in [help/DETECT-ALARM-LAUNCH-ANDROID.md](https://github.com/gdelataillade/alarm/blob/main/help/DETECT-ALARM-LAUNCH-ANDROID.md), including the `onCreate`/`onNewIntent` handling, `setIntent`, and a note that Dart `main()` hasn't run during `onCreate` so the value has to be pulled rather than pushed.

Thanks for pushing on this — the docs are better for it.

samyak1409:
Thanks — you were right, and closing it was the right call. `ACTION_RING` vs `ACTION_MAIN` is a cleaner signal than comparing extras on two otherwise identical intents, and I'd read `resolveRingIntent()` as implying a dedicated activity rather than any activity, which is what sent me toward patching the fallback.

Both my apps are on it now and the heuristic is deleted rather than tuned, which is what I'd hoped for. Two notes from doing it, in case they're worth a line in the doc:

**The filter has to go on every launcher component, not just `MainActivity`.** Both apps offer alternate launcher icons, which on Android means `activity-alias` entries and flipping component-enabled states — so when a user picks icon 2, `MainActivity` itself is `COMPONENT_ENABLED_STATE_DISABLED` and the alias is the live entry. `queryIntentActivities(intent, 0)` skips disabled components, so a `RING` filter on `MainActivity` alone silently stops resolving for anyone not on the default icon, and the app quietly goes back to the launcher-intent behaviour. Declaring it on all three works cleanly — exactly one is ever enabled, so exactly one ever matches and your multiple-handler warning stays quiet.

**Visibility, not just the action, is what "the alarm opened us" means.** With `singleTop`, an `ACTION_RING` intent also reaches an activity the user already has open — the alarm fires while they're using the app — and hiding on stop there would take away a screen they chose. Recording the launch only when the activity was not already visible (tracked across `onStart`/`onStop` rather than `onResume`/`onPause`, since a system dialog pauses without hiding) covers both cases that should hide: a cold start, and a backgrounded app brought forward.

The `onCreate`/`onNewIntent`/`setIntent` and pull-don't-push notes in the doc were exactly the parts I'd have got wrong, so thanks for writing it up properly.

Issue #418, opened 2026-08-04
[Android] A stale alarm rings immediately at boot instead of being discarded


gdelataillade:
## Summary

`BootReceiver` re-arms every stored alarm through `AlarmApiImpl.setAlarm`. For an alarm whose time passed while the device was off, `AlarmScheduler.schedule` takes its immediate branch and fires it right away:

```kotlin
val armed = if (delayInMillis <= IMMEDIATE_THRESHOLD_MILLIS) {
    Handler(Looper.getMainLooper()).postDelayed(
        { context.sendBroadcast(intent) },
        delayInMillis.coerceAtLeast(0L)   // past due -> 0 -> rings now
    )
```

Dart's reconciliation does the opposite: `_checkAlarm` stops a past-due alarm that isn't already ringing. The two paths disagree about the same input, so boot rings what `Alarm.init()` would have discarded.

## Verified

Pixel 8a / Android 16, alarm 5.8.0. Armed an alarm 30s out, rebooted immediately so it fell due mid-boot, then touched nothing — no manual launch, no unlock:

```
18:47:22  ActivityManager: Start proc ... for broadcast {.../BootReceiver}
18:47:26  Background started FGS: Allowed [... cmp=.../AlarmService; code:BOOT_COMPLETED]
```

The alarm was due at 18:46:55, 31 seconds earlier. In an earlier run with a 2.5-minute-stale alarm the same path produced a full, user-visible ring.

## Decision needed

- Discard stale alarms at boot, matching Dart's policy
- Keep ringing them
- Post a silent "missed alarm" notification instead, as platform clock apps do

I lean toward the first: two reconciliation paths disagreeing is the actual defect.

## Related observation

In the screen-off run the ring half-failed: `AlarmService` started and stayed running, but no notification was posted (`startForegroundCount:0`) and no audio played. Worth checking whether posting the foreground notification is reliable inside the boot window.

Possibly the cause of #391, though that report's specific symptom did not reproduce.

samyak1409:
One consumer's view, so weight it accordingly.

I ship two alarm apps on this package. One rings at dawn — so a phone that dies overnight and boots at 08:00 sounds today's 06:00 alarm on the spot, with nothing in the app's own history to explain it. **So: discard, yes.** Two reconciliation paths disagreeing about the same input is the defect.

[...]

Issue #424, opened 2026-08-06
[Android] An alarm due within ~45s of boot never rings: mediaPlayback FGS is refused under BOOT_COMPLETED attribution


gdelataillade:
## Summary

On Android 15+ (`targetSdk` 35+, which is Flutter's default), an alarm that fires within roughly the first 45 seconds after the device finishes booting **never rings, and nothing reports it**. `AlarmService` starts, `startForeground()` is refused, and the service is left running without a notification or audio:

```
android.app.ForegroundServiceStartNotAllowedException:
    FGS type mediaPlayback not allowed to start from BOOT_COMPLETED!
```

Android 15 restricts which foreground service types may be started from a `BOOT_COMPLETED` broadcast, and `mediaPlayback` — the type `AlarmService` declares — is on the restricted list. Crucially the restriction follows the *attribution*, not the caller: while the `BOOT_COMPLETED` temporary allowlist is active (20 s, starting when `BootReceiver` runs), **every** foreground service start by the app is attributed to `BOOT_COMPLETED`, including one triggered by an ordinary `AlarmManager` delivery.

`AlarmService.ringAlarm` catches this and returns:

```kotlin
} catch (e: ForegroundServiceStartNotAllowedException) {
    Log.e(TAG, "Foreground service start not allowed", e)
    return
}
```

so the failure is invisible to Dart: `BootReceiver` has already logged `Alarm rescheduled successfully`, and the ring dies later, in a different process context, where nothing propagates it. The #420 machinery cannot see it.

## Distinct from #418

#418 is about *policy* — whether a stale alarm should ring at boot at all. This is about a legitimately future alarm that the plugin correctly re-arms and then cannot sound. Different cause, different fix, and a wider footprint: discarding stale alarms does not address it. #418's "Related observation" (`startForegroundCount:0`) is this bug.

Plausibly also the real #391.

## Verified

Pixel 8a, Android 16 (`CP1A.260505.005`, SDK 36), alarm 5.9.0 as published, example app at `targetSdk` 36. Three runs, differing only in when the alarm falls relative to boot.

**A — alarm fell due during boot** (stale, so `AlarmScheduler` takes its immediate branch):

```
14:57:30.248 I/ActivityManager: Background started FGS: Allowed [callingPackage: com.gdelataillade.alarm.alarm_example;
             cmp=.../AlarmService; code:BOOT_COMPLETED;
[...]

Issue #419, opened 2026-08-04
[Android] checkAlarm() stops an alarm that is due but has not started ringing yet


samyak1409:
## Summary

`Alarm.init()` calls `checkAlarm()`, which stops any saved alarm whose time has passed and which the platform does not report as ringing:

```dart
final isRinging = iOS
    ? await IOSAlarm().isRinging(alarm.id)
    : await AndroidAlarm().isRinging(alarm.id);
if (isRinging) {
  _ringing.add(_ringing.value.add(alarm));
  ringStream.add(alarm);
} else {
  final current = await getAlarm(alarm.id);
  if (current != null && current.dateTime.isAfter(DateTime.now())) {
    // …left scheduled
    continue;
  }
  await stop(alarm.id);
}
```

"Fired and was already stopped" and "due, but the audio has not started yet" are indistinguishable there, and both end at `stop(alarm.id)` — which cancels the pending `AlarmManager` alarm and unsaves it.

An alarm due at 06:00:00.000 is not audible at 06:00:00.000. On Android the broadcast → `AlarmService` → `MediaPlayer` chain takes a moment, so an `Alarm.init()` landing anywhere in that gap cancels a ring that was about to sound. Nothing reports it: `set()` succeeded earlier, and the alarm simply never rings.

The snooze re-read added in 5.7.0 doesn't cover this — it only spares an alarm whose `dateTime` has moved *into the future*.

## Why it matters

The obvious case is a cold start on the alarm minute, which is rare. The one that bit us is not rare and is easy to miss: **Dart statics are per-isolate, so `Alarm.init()` runs again in every background isolate.**

Our app runs periodic background checks through `android_alarm_manager_plus`. Those run in a separate isolate, so the plugin gets initialised there too. On a normal morning an earlier check has already created that isolate and `checkAlarm()` is long past — but when the isolate's *first* `init()` lands at or after the alarm time, it runs straight on top of the pending ring:

- an alarm created shortly before it is due (save a 06:00 alarm at 05:59:30, and the 06:00 background check is that isolate's first), or
- the process being killed mid-morning (Doze, OEM task killer, low memory) and revived for a check at the alarm time.

The result is a silent morning with a successful `set()` behind it, which is the worst shape a failure can take in an alarm app.

## Suggested fix

A grace window: don't stop an alarm whose time passed moments ago, and let it fire.

```dart
final current = await getAlarm(alarm.id);
if (current != null && current.dateTime.isAfter(DateTime.now())) {
[...]

PR #431, opened 2026-08-19
feat: let the application acknowledge alarm events (#429)

### LINKED ISSUE #429: [Dart] A host alarm event is acknowledged before the app's handler finishes, so a process death loses it
Split out of #418, where it came up while wiring 5.10.0's events into two apps. It has nothing to do with stale alarms — it is the drain itself, so it applies to every verb and cause.

## Summary

In `_applyPendingEvents`, the event is handed to Dart and the durable native marker is deleted in the next statement:

```dart
if (await _applyEvent(AlarmEvent.fromWire(wire))) {
  applied.add(wire.alarmId);
}
// Acknowledged either way: an applied event is durable, and one that
// could not be applied is not going to become applicable later.
await AlarmApi().acknowledgeAlarmEvent(
  alarmId: wire.alarmId,
  recordedAtMillis: wire.recordedAtMillis,
);
```

`_applyEvent` → `_applyDrop` ends with `_events.add(event)`. A stream delivers to listeners asynchronously and **discards whatever Future the handler returns**, so `acknowledgeAlarmEvent` is not waiting for the application to have done anything with the event. Any handler that touches storage — which is every real one, since the event exists precisely so it can be written somewhere — is still in flight when the marker is deleted.

A process death in that window loses the event permanently. The native marker is gone, so no later `getPendingAlarmEvents()` returns it, and the buffer that would otherwise replay it is a `ReplaySubject`: in memory, and replaced outright in `init()`.

## Ordering, measured

Modelling the drain's exact shape — `ReplaySubject` at its default `sync: false`, an `await` between `add` and the ack, and a handler that awaits one write — gives:

```
plugin: _events.add
plugin: calling acknowledgeAlarmEvent
  app: handler STARTED
plugin: ACK DONE (native marker deleted)
  app: handler FINISHED (claim written)
```

So a *synchronous* handler is safe: it starts and completes before the ack resolves. Nothing that persists anything is synchronous. The gap is between those last two lines, and it is as long as the app's write takes — not a microtask, a disk round trip, at the point in the process lifetime where the app has just started and is doing everything else at once.

(That is a standalone reproduction of the ordering, not the plugin under test. The code references above are the real path.)

## The existing guard is defeated one layer up

`_applyDrop` already carries exactly the right instinct:

> A replay from a *different* run is deliberately reported again. Removing the alarm is durable and completes before the report, so a process death in between leaves the marker with the alarm already gone — and treating "the alarm is missing" as proof it was reported would swallow the only notice the application ever gets. A duplicate is covered by the documented `(id, recordedAt)` key; silence is not recoverable.

That reasoning is right, and it is the argument for this issue too. The cross-run replay it depends on can only happen if the process dies *before* the ack — a window of one platform-channel round trip. Once the ack lands, while the handler is still awaiting its first write, there is nothing left to replay from and the deliberate re-report can never fire.

Note the asymmetry it creates. In `_applyDrop` the alarm's *removal* is durable and ordered before the report, so it survives a death here. Only the notice is lost — which leaves the end state "the alarm is gone and the app was never told", the one outcome that comment sets out to prevent.

## What it costs a consumer

We keep our own claim key and mark it only after our handler succeeds, so re-delivery is safe on our side — but that only helps while a re-delivery is still possible, and it stops being possible the moment the marker is deleted. No amount of care in the app can recover it; the durable record is upstream of us.

For `dropped` specifically the event is the *only* evidence the alarm did not ring. Lose it and the morning shows nothing at all — not a missed alarm, just an absence the app cannot explain.

## Possible fixes

1. **Let the app acknowledge.** Expose `Alarm.acknowledgeEvent(event)` and stop acking on the app's behalf. This puts the durability boundary where the durable work happens, and the per-verb TTLs already shipped (24 h for `dropped`, a week for `moved`) bound what an app that ignores events can accumulate — so the failure mode it introduces is already handled. It is an API change, so presumably it wants a flag or a major.
2. **Ack only after a listener has consumed it** — keeps the API as-is, but "consumed" is not something a plain `Stream` can report, so it means the drain awaiting something the subscription hands back.
3. **Defer the ack by one run**, leaving the marker for the *next* `init()` to clear. Smallest change and no API surface, at the cost of one deliberate duplicate per event — which the documented `(id, recordedAt)` key already tells apps to expect.

I lean to 1, because it is the only one where the plugin stops guessing on the app's behalf, and the TTLs mean it costs nothing new. But 3 buys most of the safety for almost nothing if you would rather not move the API.

Happy to send the PR for whichever you prefer.

samyak1409:
Closes #429, on the shape you decided there: `Alarm.acknowledgeEvent(event)` plus an opt-out of the automatic acknowledgement now, manual as the only path in 6.0.0.

## What it does

`Alarm.acknowledgeEvent(event)` and `Alarm.init(acknowledgeEventsAutomatically: false)`. The default is untouched, so nobody already on 5.10.0 sees a difference.

The plugin still acknowledges every event the application was never given — one that could not be applied at all, and under the default a deferral an earlier run already applied. Nobody else can: those never reach `Alarm.events`, so no listener knows they exist, and their markers would come back on every `init` until they expired.

A repeat suppressed *because the application already has the event* is deliberately **not** in that set. Reading "we did not emit it this time" as "nobody has it" would let a second drain in one run delete the marker while the first handler was still writing — which is the same hole in a smaller window.

## One thing outside the drain, and I'd like your call on it

**The live path had the same bug, and I fixed it here rather than ship an opt-out that is a false promise.**

`AlarmService.reportHostEvent` and `SnoozeCoordinator` dropped the marker when the Dart `alarmEvent` call replied. That reply means Dart *emitted* the event, not that anything finished with it — `_alarmEvent` returns as soon as `_applyEvent` does. So an application that took the manual boundary would still have lost a live-delivered drop, silently, having been told it was covered.

Both now leave the marker for Dart, which acknowledges under the same rule as the drain. It is confined to those two files and reverts cleanly if you would rather it landed on its own.

## Two decisions worth a look

**An `init` that does not name the setting leaves it alone** — hence `bool?` rather than a defaulted `bool`. `init` is callable more than once, on resume or from a second library path, and a bare call is not an application changing its mind about durability. Letting it restore the default would hand the boundary back from somewhere that never mentioned it, and start acknowledging events the application still owns. An explicit value still moves it either way, warning when it undoes an earlier opt-out.

**A deferral whose marker outlived a restart is reported again under the opt-out.** `_applyMove` stores the moved time before it emits, so on the next launch the stored alarm already matches and `isNewDeferral` is false. That proves the *plugin* applied it, never that the application recorded it — and the surviving marker is proof it did not. Treating "already applied" as "already delivered" would silently drop exactly the event the opt-out exists to keep. This follows what `_applyDrop` already does deliberately across runs: a duplicate is covered by the documented `(id, recordedAt)` key, silence is not recoverable. Untouched under the default.

## Testing

103 Dart tests (14 new) and the 17 Android ones pass; `flutter analyze` and `dart format` are clean.

Each new guarantee was checked by deleting the fix and confirming the matching test goes red: the opt-out itself, Dart acknowledging the live path, the suppressed-repeat rule, the restart re-emit, a bare `init` not handing the boundary back, and `checkAlarm`'s single-flight.

## Two things not covered, said plainly rather than left to look covered

**The retention sweep on `_reportedEvents`.** The set now holds moves as well as drops, and a live snooze mints a new key each time, so entries older than the longest marker TTL are dropped — on emit and at the top of every drain. There is no direct test: the scenario it guards cannot arise while the host expires its own markers first, so a test would assert unreachable behaviour. It is a memory bound, not a behaviour.

**The removed native acknowledgements.** The Dart tests prove Dart now asks for the acknowledgement; they cannot prove the host stopped sending its own. Covering that means mocking a static `AlarmPlugin.alarmTriggerApi` from a private method on a `Service`, which needs a mocking dependency and some injection — more than this change should drag in, given how deliberately minimal the JVM test source set is. Say the word if you want it and I will do it separately.


gdelataillade:
Thanks for this — the reasoning is careful throughout and the tests are the best this repo has had; the boundary is in the right place. Verified locally: analyze and format clean, 103 Dart and 17 Kotlin tests pass, and the version/CHANGELOG match the additive surface. Two doc gaps worth sorting before merge, both confined to apps that take the opt-out:

- `lib/alarm.dart` — **`Alarm.snoozed` inherits the redelivery but hands out no way to end it.** Under the opt-out a surviving `moved` marker is re-emitted on each launch (your own restart test shows this), and `snoozed` is a view over it, so a snooze re-fires every launch until the marker expires a week later. But a `snoozed` subscriber gets `({int id, DateTime nextRingAt})` and `acknowledgeEvent` needs `recordedAt`, so that listener has nothing to acknowledge with. Worth saying that taking the boundary means listening to `events` instead — `example/lib/screens/home.dart:45` is the shape that would re-show a stale SnackBar on every launch.

- `lib/alarm.dart` — **the setting is per-isolate, and the docs read as though it isn't.** It is static state initialised to `true`, so an isolate that never names it starts at the default rather than at "unset"; a bare `init()` there acknowledges markers the opting-out isolate still owns. I reproduced it with `resetForTesting` (documented as "the state a fresh isolate would see") plus a bare `init`. Not a regression — 5.10.0 acknowledges everywhere already — and it needs an app-created background isolate to bite, since the plugin spawns none. So just a caveat on `init`, but "Omitting it leaves the boundary as it is" is what makes it worth spelling out.

Keeping the Kotlin change, by the way — without it the flag would be a false promise for exactly the boot-time events it matters most for, so I'd rather it landed here than separately. Both admitted test gaps are the right calls too.

Thanks again — happy to take another look once the docs are in.

samyak1409:
Both added in 05fdce1.

`snoozed` now says outright that it is not a stream the manual boundary can be built on — the subscriber gets `(id, nextRingAt)` and `acknowledgeEvent` needs `recordedAt`, so there is nothing to acknowledge with, and points such an application at `events` with a filter on the snooze cause instead.

`init` now says the setting is per-isolate: static state that starts at the default, so an isolate the application spawns does not inherit an opt-out, and a bare `init()` there would acknowledge markers the main isolate still owns. With the note that the plugin spawns none itself, so it only reaches an application that creates one.

Comments only, no logic touched. Analyze and format clean, 103 Dart tests still pass.

Thanks for keeping the Kotlin change — agreed it would have been a false promise for boot-time events otherwise.

gdelataillade:
Both of these are live in **5.11.0**, published today. Thank you — between them they closed #418 and #429 and found two things I would not have.

A few things happened after you stopped watching, so you are not guessing at them.

**I rebased your branch myself rather than waiting for you.** #432 went first, so this needed the version bump you offered to do; I merged `main` into your branch and pushed it (`6ae701c`). A merge rather than a rebase, deliberately, so nothing of yours was rewritten and no force-push landed on your fork. Sorry for moving it under you — it was to keep the release moving, not a comment on the offer.

**It shipped as 5.11.0, not 5.11.0 + 5.12.0.** 5.10.0 was the last version actually published, so two separate sections would have described a release nobody could install. They are collapsed into one entry, and trimmed hard to match this changelog's house style — 226 words down to 69. That is convention, not a verdict on the writing; the reasoning you wrote now lives in the dartdoc and a new README section, where someone deciding how to use these is already reading.

**I withdrew the 6.0.0 promise, and this one is substantive.** `Alarm.snoozed` hands its subscriber `(id, nextRingAt)`, and `acknowledgeEvent` needs `recordedAt` — so that stream, public since 5.7.0, has nothing it can acknowledge with. Making manual acknowledgement mandatory would leave it replaying every deferral on every launch for a week, with no remedy short of migrating to `Alarm.events`. That is a real flaw in the staged plan rather than a preference, so automatic acknowledgement stays the default and the manual boundary is a permanent advanced opt-in. Your `05fdce1e` doc note is what made it obvious.

**Two small changes to your code**, both on `main`: `androidStaleAfter` needed a row in the README settings table, and `AlarmSettings.kt`'s lenient parser threw on a non-primitive rather than recovering — `stale.jsonPrimitive` raises on an object or an array, so one malformed field cost the whole alarm instead of falling back to the default like your Dart reader does. Now `(stale as? JsonPrimitive)?.content?.toLongOrNull()`, with a test over `{}`, `[]` and a populated object; reverting the cast fails it alone.

**Your marker-expiry find was the most valuable part of #432** and I want to say so explicitly: ageing from `atMillis` would have made every `STALE_AT_BOOT` marker expire before Dart could read it, so the feature would have shipped silently inert. I confirmed it by reverting `maxOf` — exactly the two failures you described, with the pruning test still passing.

**Verified on a Pixel 8a before publishing**, since nothing about the boot path is reachable from unit tests:

- stale alarm discarded, not rung: `Alarm 321 was due at 13:06:00 and is past its stale window; discarded instead of ringing at boot`, with `{"alarmId":321,"verb":"DROPPED","cause":"STALE_AT_BOOT",…}` written
- that marker drained on next launch as `AlarmDropped id=321 cause=staleAtBoot`, then cleared
- a *non*-stale alarm re-armed, refused inside the boot window, deferred 30 s by #424, and rang audibly at the retry
- `"androidStaleAfterMillis":null` stored as null, and still not discarded with a 1 s cutoff forced

Also worth knowing: the default is *absent* from stored JSON, exactly as your `an alarm stored before the cutoff existed takes the default` test predicts, so the upgrade path and a fresh default are the same shape on disk.

Two gaps left open, if either interests you. `_checkAlarm` still stops an equally stale alarm silently when there was no reboot, so the same situation reports an event or nothing depending on how it was found — it wants its own cause. And `Alarm.snoozed` still has no acknowledge handle, which is now a documented dead end rather than a blocker.

samyak1409:
Thanks for writing all that up — and no apology needed for the rebase. Merging `main` in rather than rewriting the branch was the considerate version of it, and the release moving is worth more than my doing the bump myself.

The changelog collapse is obviously right: two sections describing a release nobody could install is worse than a trim, and the reasoning reads better in the dartdoc than in a changelog entry anyway.

**The Kotlin parser was my miss, and a specific kind of one.** I spent real effort making the Dart reader recover from every malformed shape — non-numbers, non-finite, fractions, unrecognised negatives — and then wrote the Kotlin fallback as though the only failure mode were "does not parse as a Long". `jsonPrimitive` throwing on an object or an array is the same class of bug I had just finished designing around, and since `getSavedAlarms` catches per alarm it cost strictly more there than the equivalent would have in Dart. I should have carried the leniency across instead of re-deriving it. Same story for `'$_ringStartGrace'`: I wrote that message and never read it back.

Thanks for the Pixel run too — the refused-then-deferred-then-rang sequence is the one I most wanted to see and had no way to.

Both gaps interest me. Taking them in the order you listed, with the parts I think are yours to decide.

## `_checkAlarm`'s silent stop

Agreed it wants its own cause: `staleAtBoot` means "the device was off and this is too late to be a wake-up", and this one means "it came due, the platform is not ringing it, and the grace has passed" — a different question with a different window. Three things I would rather settle before writing it.

**Durability, which I think is the real decision.** Every event on `events` today is born as a durable host marker, and that is what `acknowledgeEvent` is keyed on. An event created in Dart inside `_checkAlarm` has no marker, so under `acknowledgeEventsAutomatically: false` it quietly opts out of the guarantee the flag exists to provide — a process death between emit and the app's write loses it with no replay, exactly the failure the boundary was added to close. Two ways out:

- emit it from Dart and document it as the one event on the stream that is not durable; or
- have Dart ask the host to record a marker first and drain it in the same pass, the way a native drop already works.

I lean to the second, for the reason `AlarmDropped` is durable in the first place: this event is the only evidence the alarm did not ring. The cost is very uneven across the two platforms though, and that is worth naming before you weigh it. Android has a marker store to write into. iOS has none at all — `getPendingAlarmEvents` returns `[]` and `acknowledgeAlarmEvent` is a deliberate no-op, so no `AlarmEventWire` has ever been produced there. Making this durable on iOS would not be wiring into an existing mechanism, it would be building the first one, which is a much larger ask than the Android half and tilts me further towards the Android-only option below than I was when I started writing this. The Dart-only version is a few lines on both.

**iOS reaches that branch differently.** `stopAll()` runs before the loop, so by the time an alarm gets there `getAlarm` is null and `isRinging` cannot come back true — every past-due alarm takes the stop path, with no grace applied. Ring state is in-memory on iOS, so on an ordinary cold launch nothing was audible to lose; the case that stings is a second `init` in a live process. On Android the 30 s grace and the `isRinging` check make the same `stop()` mean something much narrower. Emitting from that line as it stands would report a drop on iOS for alarms Android would spare. Options as I see them: land it Android-only and say so, or fix the iOS ordering first as its own change. I would rather not quietly ship an event that means two different things per platform. Worth knowing either way that the current iOS behaviour is pinned, not drifting — `iOS still stops a past-due alarm` in `test/alarm_check_test.dart` asserts it deliberately, and I wrote that test in #432 to record the behaviour rather than to endorse it. So changing it means rewriting an assertion on purpose, which is another reason I think it wants its own issue rather than riding along on this one.

**The name.** `staleAtCheck` reads as a sibling of `staleAtBoot` and points at `checkAlarm`, which is public and discoverable. `missedAtCheck` is the more honest description of what was observed and does not borrow "stale" for a different window. Slight preference for the second; happy with whichever you pick.

## `Alarm.snoozed`

The narrow fix is to stop presenting it as an option: `@Deprecated`, pointing at `events` with a filter on `AlarmEventCause.snooze`. It is now strictly a lossy view of a stream that does everything it does, and the loss is exactly the field the boundary needs. `updateStream` is already deprecated in that file with no removal version attached, so there is precedent for doing this without owing anyone a 6.0.0.

The alternative, if you would rather it stayed usable, is to widen the record to `({int id, DateTime nextRingAt, DateTime recordedAt})` and add an id-and-timestamp overload of `acknowledgeEvent`. That is source-compatible for any ordinary `listen` callback, but it grows the surface of a stream you would rather people left, so I would not push for it.

Happy to open a PR per gap once you have picked. If you would rather spec these in an issue first, as with #418, that works too — that shape worked well.

gdelataillade:
Decisions below, and one correction first.

**`iOS still stops a past-due alarm` is ours, not #432's.** It came in with `d50a39e`, the #419/#420 fix. Your reading of it is exactly right — it was written to *record* that behaviour, not to endorse it, and the plan that produced it said so in as many words — but the intent you're crediting yourself with was mine, and I'd rather you knew than found out later.

On the parser: agreed, and I'd put it more kindly than you did. You designed the leniency carefully in one language and then wrote the other half as if the failure set were smaller. That is the most ordinary way for a good design to leak, and it is exactly what a review is for.

## `_checkAlarm` — all four calls

**Durability: your option (b), a host-recorded marker.** `events` documents at-least-once with a durable marker behind every member, and one member that quietly opts out makes the guarantee non-uniform — worse than not having the event. Your objection was the uneven platform cost, and I think that dissolves under the next decision.

**Android-only, explicitly.** Your iOS analysis is correct and it settles this: `stopAll()` at [alarm.dart:348](../blob/main/lib/alarm.dart#L348) runs before the loop, so on iOS every past-due alarm reaches [:397](../blob/main/lib/alarm.dart#L397) with `getAlarm` null, no grace applied and no ring state to consult. Emitting from there would report a drop for alarms Android deliberately spares — one event name meaning two different things. Android-only means iOS emits nothing, needs no marker store, and the ordering question stops blocking this. Filed separately as the iOS issue below; the pinned test stays as it is.

**`missedAtCheck`.** Your reasoning wins: `staleAtBoot` answers "the device was off and this is too late to wake anyone", this answers "it came due and the platform never rang it". Different window, different question — borrowing "stale" is something we would have to un-teach.

**Two things to fold in that you didn't raise.** `Alarm.events`' own doc says it reports "deferrals and discards the host made on its own", which stops being strictly true once Dart initiates one — that wants widening deliberately rather than quietly meaning something new. And I would keep the new host surface narrow: `recordMissedAlarm(alarmId, scheduledForMillis)` rather than a general "record this event", so verb and cause stay the host's decision and Dart cannot mint arbitrary markers.

## `Alarm.snoozed`

**Deprecate, don't widen.** It is now strictly a lossy view of `events` and the missing field is precisely the one the boundary needs. Widening the record would grow the surface of a stream we want people to leave. `updateStream` and `ringStream` are the precedent you found, and neither carries a removal version.

## Priority, honestly

The `_checkAlarm` gap is a consistency wart *I* raised, not something a user has reported, and it now costs a pigeon method plus native work. The deprecation is a few lines with immediate value. If you only do one, do the deprecation.

Both filed as issues so you have something to build against rather than reconstructing it from this thread: the spec, and the iOS ordering as its own. Take them in whichever order suits — and no obligation on either; you have already done more than your share of this release.

gdelataillade:
Filed: #433 (the `missedAtCheck` spec, with the four decisions above baked in), #434 (the iOS `stopAll()` ordering it depends on), and #435 (the `Alarm.snoozed` deprecation). #435 is the one I'd start with if you only pick one.

samyak1409:
Thank you for the correction, and for making it directly. You are right: `d50a39e` is yours, and `git log -S` on the test name confirms it — the assertion arrived with the #419/#420 fix, not with #432. I had reconstructed authorship from the reasoning rather than checking, which is the easy way to end up sincerely wrong, and it is worth my noticing that I did it in the direction that flattered me. The reading was right because you had already written the reason down; I just found it.

All four decisions make sense, and the pair of them resolves the thing I could not settle. Android-only is what makes the durable marker affordable — the objection I raised was entirely about the iOS half, and removing it removes the objection rather than overriding it. `recordMissedAlarm(alarmId, scheduledForMillis)` over a general recorder is better than what I had in mind: I was thinking about where the marker is written, not about who gets to decide what it says, and a general "record this event" would have handed Dart a pen it has no business holding. Both points you folded in are ones I would have missed — the `Alarm.events` doc sentence in particular, since "the host made on its own" is exactly the kind of phrase that quietly stops being true.

#435 is up as #436, since it was the one you said to start with and I agree with the ordering. It is the deprecation and nothing else, plus the example and README moved off the deprecated call — the example was demonstrating the precise failure the issue describes, a snooze snackbar a manual-boundary app could never acknowledge.

Two tests came with it that are not strictly about the deprecation. Once the README tells applications to write the `AlarmMoved` + `snooze` filter themselves, that advice is only correct while the filter and the getter agree, so I pinned the equivalence rather than leaving it to a one-line body. I checked both against a mutated getter rather than trusting them to bite.

On the other two: no argument with your priority read, and I am not picking up either right now. Both are specified well enough in the issues to stand on their own, so nothing is waiting on this thread. On #434 in particular, my earlier analysis was about what the ordering does, and I would not want it read as a plan to change it — that one alters whether an alarm rings, partly on a platform where some of the verification would be against a simulator, and it deserves more certainty than I have from reading the code.

PR #432, opened 2026-08-20
feat(android): discard alarms that went stale while the device was off

### LINKED ISSUE #418: [Android] A stale alarm rings immediately at boot instead of being discarded

samyak1409:
Closes #418. Built to the specification in your last comment there, with the serialization change from the comment above this one.

### What changed

`BootReceiver` re-armed every stored alarm with no age limit. It now discards one missed by more than `AlarmSettings.androidStaleAfter` and records a `DROPPED` marker with cause `STALE_AT_BOOT`. Nothing is reported from there — no engine is running at boot — so Dart drains it on the next `init()` and it surfaces through `Alarm.events`. The plugin still shows the user nothing itself.

`AlarmSettings.androidStaleAfter` defaults to 15 minutes; an explicit `null` never discards.

### The one place I departed from the spec

`Duration?` with a 15-minute default is unchanged, but it needed help to persist. `json_serializable` omits a null field from the encoded map and reads a missing key as the constructor default, so "never discard" became 15 minutes on the first launch after it was set. `androidStaleAfter` is written as a `-1` sentinel instead of JSON null, which keeps absent and null apart without touching the public shape. Since `toJson()` is public, the sentinel is documented on the field as part of its contract.

An unusable value — a string, a fraction, a non-finite number, any other negative — recovers to the default with a warning rather than throwing. Nothing between `Alarm.init` and the storage read catches, so a throw would cost the user every stored alarm over one malformed field, which is a worse failure than the one it would be guarding against. The Kotlin side keeps the same three states apart: kotlinx already separates absent from null, and the lenient parser checks key presence rather than relying on `primitiveLong`, which returns null for both.

`primitiveLong` staying as it is, by the way, is deliberate — the collapse only matters for a key that can legitimately be null, and changing the helper would touch every field that uses it.

### Why `AlarmStorage` is in this diff

Marker expiry aged from `atMillis`, the moment the event describes. That is right for a `MOVED`, where it is the time the alarm now rings. A `STALE_AT_BOOT` discard describes a due time that had already passed — days ago, since that is exactly what made the alarm stale — so the marker was written already older than the 24 h discard TTL and `getPendingAlarmEvents` deleted it on the first read. The device was off for a day, every stale alarm was discarded correctly, and the application heard about none of them.

It is invisible today because the only existing `DROPPED` is a refused ring, recorded within seconds of the time it describes, so there the two timestamps are interchangeable.

Expiry now ages from `maxOf(atMillis, recordedAtMillis)`. Deferrals are unaffected, since the ring time is the later value, and the change can only extend a marker's life — never shorten it — so nothing that survives today starts expiring.

### Validation

`alarmSettingsValidation` throws `invalidArguments` below the 30 s ring-start grace, at `set()` rather than silently at the next boot.

### Tests

99 Dart, 32 Kotlin (15 new), all passing; analyze and format clean.

Both sides cover the three states — absent takes the default, `-1`/explicit null never discards, a value is itself — plus the recovery path, the boundary (exactly 15 minutes still rings), a longer per-alarm window, and that discarding leaves the alarm gone with a marker naming the cause and the time it should have rung. On expiry: an alarm missed by three days still reaches Dart end to end, a discard recorded now survives a long-past due time, and a discard recorded two days ago is still pruned so the TTL keeps biting.

Revert-proved, each against the change it covers: removing the `@JsonKey` converters fails the round-trip and recovery tests, making `isStale` always return false fails two boot tests, and restoring the old expiry expression fails two of the three above.

**Not covered:** the wiring inside `rescheduleAlarms` that calls them. `BootReceiver` needs a `Context`, and the Kotlin test source set is plain JUnit with no Robolectric, so `isStale` and `discardStaleAlarm` moved to the companion object and are tested directly. The three lines that call them in the loop are not.

### Two things for you

The version bump assumes this lands before #431 — both claim 5.11.0, so whichever is second needs a rebase to 5.12.0. I merged the two locally to see what that costs: only the CHANGELOG conflicts, `lib/alarm.dart` merges cleanly despite both branches working in it, and the combined tree is green — 113 Dart and 32 Kotlin, analyze clean. Happy to do the rebase on this one whichever order you pick.

The `init()`/boot asymmetry you asked me to leave alone is untouched. `_checkAlarm` still stops an equally stale alarm silently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.11.0
+* **[Android] An alarm whose time passed while the device was off is now discarded at boot instead of ringing immediately (#418).** Previously every stored alarm was re-armed with no age limit, so an alarm set for 06:00 sounded at 08:00 when the phone was switched on. Alarms missed by more than `AlarmSettings.androidStaleAfter` are dropped and reported through `Alarm.events` as an `AlarmDropped` with cause `staleAtBoot`, so the application can tell its user the alarm was missed.
+* Added `AlarmSettings.androidStaleAfter`, defaulting to 15 minutes — long enough that a reboot straddling the alarm still rings, short enough that a phone switched on hours later stays quiet. Pass `null` to never discard, which is the behavior of 5.10.0 and earlier.
+
 ## 5.10.0
 * [Android] Fixed an alarm due within ~45s of a reboot never ringing (#424).
 * Added `Alarm.events`, reporting deferrals and discards the host made on its own.
--- a/android/src/main/kotlin/com/gdelataillade/alarm/alarm/BootReceiver.kt
+++ b/android/src/main/kotlin/com/gdelataillade/alarm/alarm/BootReceiver.kt
@@ -4,12 +4,57 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.util.Log
+import com.gdelataillade.alarm.models.AlarmEventCause
+import com.gdelataillade.alarm.models.AlarmEventVerb
+import com.gdelataillade.alarm.models.AlarmSettings
+import com.gdelataillade.alarm.models.HostAlarmEvent
 import com.gdelataillade.alarm.services.AlarmStorage
 import com.gdelataillade.alarm.api.AlarmApiImpl
 
 class BootReceiver : BroadcastReceiver() {
     companion object {
         private const val TAG = "BootReceiver"
+
+        /**
+         * Whether [alarm] was due so long before [now] that ringing it would be
+         * noise rather than a wake-up.
+         *
+         * A null `androidStaleAfterMillis` never goes stale, which is what an
+         * application asks for when the alarm has to ring however late.
+         */
+        internal fun isStale(alarm: AlarmSettings, now: Long): Boolean {
+            val staleAfterMillis = alarm.androidStaleAfterMillis ?: return false
+            return now - alarm.dateTime.time > staleAfterMillis
+        }
+
+        /**
+         * Discards [alarm] and records why, instead of sounding it hours late.
+         *
+         * Nothing is reported to Dart from here: no Flutter engine is running at
+         * boot, which is the whole reason these decisions are recorded durably.
+         * The marker is drained on the next `Alarm.init()`.
+         */
+        internal fun discardStaleAlarm(alarm: AlarmSettings, storage: AlarmStorage, now: Long) {
+            val event = HostAlarmEvent(
+                alarmId = alarm.id,
+                verb = AlarmEventVerb.DROPPED,
+                cause = AlarmEventCause.STALE_AT_BOOT,
+                atMillis = alarm.dateTime.time,
+                recordedAtMillis = now,
+            )
+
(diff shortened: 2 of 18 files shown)


samyak1409:
Found a real bug in this while reviewing it again — pushed 4b81d53.

Marker expiry in `getPendingAlarmEvents` ages from `atMillis`. For a `MOVED` that is the time it now rings, which is what should keep it alive. For a `STALE_AT_BOOT` drop it is the due time that had already passed, days ago — that is what made the alarm stale — so the marker was written already older than the 24 h discard TTL and was deleted on the first read.

Which broke exactly the promise this PR makes. Phone off for a day, every stale alarm discarded correctly, application told about none of them. It went unnoticed because the only existing `DROPPED` is a refused ring, recorded within seconds of the time it describes, so there `atMillis` and `recordedAtMillis` are interchangeable.

Expiry now ages from `maxOf(atMillis, recordedAtMillis)`. A deferral is unaffected because its ring time is the later value, and a discard ages from when it was recorded. The change can only extend a marker's life, so nothing that survives today starts expiring — `a deferral survives longer than a discard` and `an ancient deferral is pruned` still pass untouched.

Three tests added, all revert-proved against the old expression: an alarm missed by three days still reaches Dart end to end, a discard recorded now survives a long-past due time, and a discard recorded two days ago is still pruned so the TTL keeps biting. 32 Kotlin tests now, 99 Dart, analyze and format clean.

Issue #442, opened 2026-09-05
[Android] There is no durable record that an alarm rang, so a host that logs outcomes cannot distinguish a ring that happened from one that vanished


samyak1409:
When a ring starts, `AlarmService.ringAlarm` announces it straight into Dart:

```kotlin
AlarmPlugin.alarmTriggerApi?.alarmRang(id.toLong()) { ... }
```

If no Flutter engine is attached, `alarmTriggerApi` is null and the announcement is dropped. Nothing records that it was owed, and it is never replayed. Compare the two host-initiated verbs, `MOVED` and `DROPPED`, which are written to disk natively at the moment they happen and drained by Dart whenever it next runs.

## Why the gap is visible from the app side

An alarm leaves the platform once it has fired and been stopped. So for an app that reconciles after the fact, "it rang and the user stopped it" and "it was owed and simply never happened" are the same observation: no alarm, no event, nothing.

That is reachable without anything going wrong. `androidFullScreenIntent` usually brings the app up, but Android is explicit that it might not — "While the user is using the device, the system UI might display a heads-up notification instead of launching your full-screen intent". The user taps Stop on that notification, the app never comes to the foreground, Dart never runs, and the only record of a successful ring was spoken to nobody.

For us that surfaces as an outcome in a permanent log — a genuine ring the user woke up to gets written down as unconfirmed, beside the wind readings that decided it. There is no way to do better from outside the package, because the missing fact only ever existed inside it.

## The ask

A durable marker for "this alarm rang", with the same lifecycle the existing verbs already have: written natively at ring start, surviving process death, drained and acknowledged by Dart.

**Three things in `AlarmStorage` decide whether that works, and two of them would silently defeat it:**

1. **`unsaveAlarm`'s keep-list is `DROPPED`-only.** It ends `if (verb != AlarmEventVerb.DROPPED) preferences.remove(eventKey)`, and stopping an alarm reaches it — `AlarmReceiver`'s stop action calls `handleStopAlarmCommand` → `unsaveAlarm`, and so does `AlarmApiImpl.stopAlarm`. So a `RANG` marker written at ring start is deleted the moment the user presses Stop, which is precisely the sequence this issue exists to record. The comment there reasons about deferrals-versus-discards, which is the right rule for the verbs that exist today; a delivery fact is a third category and has to be added to that list deliberately.
[...]

Issue #441, opened 2026-09-05
[Android] Alarm.set on an existing id cancels and recreates the alarm, though the PendingIntent is already built for an in-place replace


samyak1409:
An alarm's notification text travels inside the alarm itself — `AlarmScheduler.schedule` serialises the whole `AlarmSettings` into the broadcast intent:

```kotlin
val intent = Intent(context, AlarmReceiver::class.java).apply {
    putExtra("id", alarm.id)
    putExtra("alarmSettings", Json.encodeToString(alarm))
}
```

So changing what a *pending* alarm will say when it fires means calling `Alarm.set` again with the same id. That is the only door, and it is a destructive one.

## What the replace costs today

`Alarm.set` stops the existing alarm first, in Dart:

```dart
for (final alarm in alarms) {
  if (sameId || shouldReplaceSameSecond) await Alarm.stop(alarm.id);
}
```

and again natively, in `AlarmApiImpl.setAlarm`:

```kotlin
if (alarmIds.contains(alarm.id)) {
    stopAlarm(alarm.id.toLong()) {}
}
```

`stopAlarm` calls `cancelPendingBroadcast`, which cancels the `AlarmManager` entry. So between the stop and the new `schedule` the alarm **does not exist**: a process death in that window leaves nothing armed, and nothing to notice it — from Dart's side `set` simply never returned.

## The platform does not require this

`AlarmScheduler` already builds its PendingIntent for exactly this case:

```kotlin
PendingIntent.getBroadcast(
    context, id, intent,
    PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
)
```

`FLAG_UPDATE_CURRENT` updates the extras of the existing PendingIntent, and arming `AlarmManager` with it replaces the existing alarm rather than adding one — atomically, payload included. The destroy-and-recreate window is the plugin's, not the platform's.

## Why it matters to a consumer

We run weather-conditional alarms. An alarm scheduled for 06:00 is decided repeatedly as the morning approaches, and each check refreshes the numbers the ring will show — the wind reading and the time it was taken. Nothing about the alarm moves; only the text does.

Today every one of those refreshes cancels a good alarm and builds a new one. It works, and it has never visibly failed, but the alarm spends a moment not existing several times per morning for a change that never needed to touch the schedule at all. It also drags the alarm through `schedule`'s near-term branch when a refresh happens close to the alarm's own minute, which is a separate problem filed as #440.

## The ask

[...]

Issue #437, opened 2026-09-03
Fatal NPE in VolumeService.startVolumeEnforcement makes alarms unstoppable in production (plus two related gaps)


fh-arnob:
Three findings from taking `alarm` 5.10.0 into production in a medication
reminder app (Android). The first is a crash that reached real users; the
other two are gaps I hit while diagnosing it. Happy to split into separate
issues if you'd prefer.

---

## 1. Fatal NPE in `VolumeService.startVolumeEnforcement`, reached whenever an enforced-volume alarm is stopped

`volumeEnforced: true` starts a runnable that re-posts itself every second by
dereferencing its own field:

```kotlin
// VolumeService.kt:44-56
volumeCheckRunnable = Runnable {
    ...
    handler.postDelayed(volumeCheckRunnable!!, 1000)   // line 54
}
handler.post(volumeCheckRunnable!!)
```

`stopVolumeEnforcement()` does `removeCallbacks(it)` and then sets
`volumeCheckRunnable = null`. It is reached from `restorePreviousVolume()`,
which runs when the alarm stops.

If the runnable is **already executing** on the main thread when that null
lands, `removeCallbacks` cannot un-run it, and it reaches line 54 and
dereferences null. The NPE is uncaught on the main thread, so it kills the
process.

Because the trigger is "the alarm stops", the crash fires precisely when the
user tries to silence the alarm, and it hits roughly half the time — it races
the 1000ms cadence.

**What made it severe rather than annoying:** `ringAlarm` returns
`START_STICKY`, so Android restarts the service, while the restarted process
has no Flutter engine and therefore no Dart state able to address the alarm.
From the user's side the alarm becomes unstoppable — tapping the notification
opened the app without silencing it, and repeated attempts each crashed
again. Affected users escaped only by rebooting or uninstalling.

Crashlytics, 16 devices, Samsung SM-S921E among them:

```
Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
       at com.gdelataillade.alarm.services.VolumeService.startVolumeEnforcement$lambda$0(VolumeService.kt:54)
       at android.os.Handler.handleCallback(Handler.java:1070)
       at android.os.Handler.dispatchMessage(Handler.java:125)
       at android.os.Looper.loopOnce(Looper.java:288)
       at android.os.Looper.loop(Looper.java:392)
[...]

Issue #384, opened 2025-09-15
Alarm.stop is not working properly in the alarm.ringing.listen stream.


cho-4:
**Alarm plugin version**
5.1.5

**Describe the bug**
```Dart
bool _wakeUpHandled = false;
  
  void initAlarmListener() {
    Alarm.ringing.listen((AlarmSet alarmSet) async {
      for (final alarm in alarmSet.alarms) {
        if (alarm.id == 42 && !_wakeUpHandled) {
          _wakeUpHandled = true;
          final isPlaying = await Alarm.isRinging(1);
          if (isPlaying) {
            await Future.delayed(const Duration(seconds: 3));
            await Alarm.stop(1);
          }
        }
      }
    });
  }
```

Here's the code I'm using. I expected alarm 1 to stop when alarm 42 rings, but alarm.stop stops all running alarms. I tried changing the id value just in case, and it still works. It stops all running alarms.

My phone is a Galaxy A34.

I'll wait for your reply. Thank you.

PR #443, opened 2026-09-10
fix(android): the volumeEnforced crash (#437) and a same-id set deleting its own Dart record

gdelataillade:
Two verified Android defects from reviewing #437–#442. Both are bug fixes with **no public Dart API change** → 5.12.1.

Each was reproduced on a Pixel 8a *before* the fix, so the repro is known-good rather than assumed.

## 1. The `volumeEnforced` crash (#437)

`volumeEnforced: true` starts a runnable that re-posts itself by dereferencing its own nullable field, and `stopVolumeEnforcement` nulls that field. samyak's Crashlytics data shows it killing the process on 16 devices at the moment the user stops the alarm; `START_STICKY` then restarts the service with no Flutter engine, so the alarm becomes unstoppable.

**The mechanism needs no second thread**, contrary to the issue body — `AlarmService` creates one `VolumeService` and every alarm reuses it, so `startVolumeEnforcement` was overwriting `volumeCheckRunnable` without cancelling the previous one. `stopVolumeEnforcement` only removes whatever is *in the field*, so an earlier runnable stayed queued and unreachable. (samyak reached the same conclusion in the issue's comment.)

Reproduced by queueing five enforced alarms and stopping them in rapid succession — each promotion orphans the previous runnable, and the final stop nulls the field:

```
AlarmService: Triggering queued alarm 9002.   (…9003, 9004, 9005)
AndroidRuntime: FATAL EXCEPTION: main
java.lang.NullPointerException
    at com.gdelataillade.alarm.services.VolumeService.startVolumeEnforcement$lambda$0(VolumeService.kt:54)
```

Same method, same line, same lambda as the field trace. After the fix, the identical procedure survives with the same pid.

**Three changes, not alternatives:** cancel before starting (the actual defect), re-post `this` rather than the field (makes the dereference impossible), and return early when the field no longer points at `this` (stops a stale runnable continuing). Note the reporter's headline fix — capturing the runnable in a local — would have traded the crash for a permanently-leaked runnable.

**Deliberately not fixed here:** `setVolume` also re-captures `previousVolume` on every call, so overlapping alarms never restore the user's original level. The obvious fix is wrong, because `activeStream` is used to compute the target and not only to restore, so it needs a saved `(stream, level)` pair and its own change.

## 2. A same-id `Alarm.set` deleted its own Dart record

Found during review, not filed as an issue.

`AlarmService.unsaveAlarm` — the notification-Stop path — clears storage and notifies Dart but cannot reach `AlarmApiImpl.alarmIds`, which is private and per-engine. It goes stale for the life of the engine, so the next same-id `Alarm.set` takes the replace preamble, which called the **public** `stopAlarm` and announced `alarmStopped` to Flutter. Dart's handler unsaves unconditionally, so that event deleted the alarm `set` had just saved.

Reproduced on device — arm, notification-stop, re-set the same id:

```
AlarmApiImpl: Stopping alarm with identical ID=9001 before scheduling a new one.
flutter: Alarm with id 9001 stopped.                      ← spurious
flutter: Alarm with id 9001 scheduled successfully at 09:40:34
```

Dart storage empty, native storage still holding `__alarm_id__9001`. Then on the ring:

```
[SEVERE] Alarm with id 9001 started ringing but the settings object could not be found.
         Please report this issue at: https://github.com/gdelataillade/alarm/issues
```

So the plugin was asking users to report its own defect. **#330, #384 and #364 look like field instances** and are worth re-reading after this lands.

After the fix, same sequence: no stop event, both stores hold the alarm, and the ring logs `Alarm with id 9001 started ringing.`

The replace path still cancels the pending broadcast (`FLAG_UPDATE_CURRENT` means a surviving entry would fire at the old time with the new settings) and still unsaves — that is the only thing clearing a pending `MOVED` marker, and skipping it would let a snooze marker survive the replace and rewrite the new time back to the deferred one, which is the trap #441 walks into.

`silenceForSnooze` becomes `silenceRing`: two callers now want "stop the audio and tell nobody".

`alarmIds` is left as it is — with the preamble silent, a stale entry costs only a redundant unsave that `AlarmScheduler.schedule` immediately overwrites.

## Testing

**No new automated tests, and that is a real gap rather than an oversight.** `android/build.gradle` declares `junit` only: `VolumeService` needs a `Context`, an `AudioManager` and a live `Looper`, and the preamble fix is about a Pigeon notification from a `Context`-constructed `AlarmApiImpl`. The Dart suite cannot see either, since both are native-side and `FakeHost` mocks the host. Covering them means Robolectric, which is a bigger decision than these fixes.

Regression gate: `flutter analyze` clean, `dart format` clean, **115 Dart** and **33 Kotlin** tests passing.


gdelataillade:
Filed the enforcement leak as #444, rather than fixing it here.

It is pre-existing: before this PR `startVolumeEnforcement` had no cancel at all, so a leaked predecessor runnable re-posted itself identically. This change only *adds* a cancel on the path where the successor also enforces, which narrows the leak without closing it — the cancel lives on a path the successor may never reach.

#444 covers it together with the two defects that share its root cause (`previousVolume` being re-captured per ring, and why `activeStream` cannot simply be frozen to fix that), since all three are per-ring volume state stored per-service and fixing them separately means three patches to the same twenty lines. Worth doing before #438, which multiplies how often the leak is reachable.

The duplicate-id finding was a real regression from this PR and is fixed in d9b896a — `alarmIds` is now a set, so `add` is idempotent and the missing `remove` cannot bite again through another path.

PR #445, opened 2026-09-10
fix(android): keep volume state per ring rather than per service (#444)

### LINKED ISSUE #444: [Android] VolumeService keeps per-ring state per-service, so enforcement leaks across alarms and the user's volume is never restored
`VolumeService` is constructed once per `AlarmService` ([AlarmService.kt:114](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt#L114)), so all four of its mutable fields — `previousVolume`, `targetVolume`, `volumeCheckRunnable`, `activeStream` ([VolumeService.kt:18-24](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/services/VolumeService.kt#L18-L24)) — are shared by every alarm rung on that instance. They describe one ring, but they outlive it.

Three separate defects fall out of that, and fixing them one at a time means three patches to the same twenty lines.

### 1. Enforcement survives into an alarm that never asked for it

`setVolume` starts enforcement only when the *new* alarm sets `volumeEnforced` ([VolumeService.kt:37](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/services/VolumeService.kt#L37)), and `ringAlarm` calls `setVolume` at all only when the new alarm has an explicit volume ([AlarmService.kt:226](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt#L226)).

On queue promotion the predecessor's cleanup is deliberately skipped while the successor starts: `restorePreviousVolume` sits inside `if (ringingAlarmIds.isEmpty())` ([AlarmService.kt:610-615](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/alarm/AlarmService.kt#L610-L615)), which is false once the promoted alarm is playing.

So if the successor has `volumeEnforced: false`, or no explicit volume, nothing stops the predecessor's runnable and it goes on forcing the *predecessor's* `targetVolume` on the *predecessor's* `activeStream` for the whole of the successor's ring. The user hears an alarm held at a level and possibly on a stream they did not ask for, and the volume rocker fights back once a second.

Bounded rather than permanent — the last stop reaches `restorePreviousVolume` and clears it — but it lasts as long as the successor rings.

### 2. Overlapping alarms never restore the user's original volume

`setVolume` re-captures `previousVolume` on every call ([VolumeService.kt:29](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/services/VolumeService.kt#L29)), so a second alarm records the level the *first* alarm had already forced. Restoring afterwards returns the device to that, not to what the user had set. With `allowAlarmOverlap` or a queue, the user's own volume is lost for good.

### 3. `activeStream` cannot be frozen to fix (2)

Worth stating because it is the trap in the obvious patch: `activeStream` is not only restore state. It is used to compute the target ([:27](https://github.com/gdelataillade/alarm/blob/main/android/src/main/kotlin/com/gdelataillade/alarm/services/VolumeService.kt#L27) and the `getStreamMaxVolume` call below it) and read by the enforcement runnable. Keeping the first ring's value so restore works would send a second alarm's target to the previous alarm's stream while its audio plays on the other one.

### Shape

Per-ring enforcement state, torn down at each transition rather than when the successor happens to want its own — and a saved `(stream, level)` pair for restore, kept separate from the stream enforcement is currently acting on. That closes all three at once.

### Not this

The `volumeCheckRunnable` NPE from #437 is a different bug and is fixed in #443; that fix narrows (1) — it now cancels when the successor *does* enforce — but cannot close it, because the cancel lives on a path the successor may never reach.

Worth doing before #438, which makes enforcement run for alarms that pass no volume at all and so multiplies how often (1) is reachable.

Found while reviewing #443. Not reproduced on a device: traced from source, and (1) in particular deserves a device check before anyone writes the fix.

gdelataillade:
Closes #444. **Draft: the device verification is still outstanding** — see below.

`VolumeService` is created once per `AlarmService`, so its four mutable fields described a single ring but outlived it. #444 traces three defects to that; this fixes all three, since they are the same bug and separate patches would collide in the same twenty lines.

## What changed

**Restore state is now a saved `(stream, level)` pair**, captured only when nothing is saved yet, and restored to *that* stream rather than whichever one the last ring happened to pick. Previously `previousVolume` was re-captured on every `setVolume`, so a second alarm recorded the level the first had already forced and the user's own level was gone for good.

**Enforcement now stops at every transition**, not only when the successor wants its own. Two places, because there are two ways in:
- `setVolume` stops unconditionally before deciding whether to start.
- `ringAlarm` stops as well, since an alarm with no explicit volume never reaches `setVolume` — that was the case @gdelataillade's review caught on #443.

The queue promotion path is what made this visible: it deliberately skips `restorePreviousVolume` while the successor starts, so nothing else was going to stop the predecessor's runnable.

**#437's protection is unchanged.** `startVolumeEnforcement` still has exactly one call site and that caller still cancels first, so no runnable is orphaned. The identity check and re-posting `this` are untouched.

## Version

Relabels the unpublished 5.12.1 as **5.13.0**. It never shipped, and this is a behaviour change rather than a pure fix, so accumulating the rest of the batch under a patch heading would have been wrong.

## Testing — read this before merging

`flutter analyze` clean, `dart format` clean, **115 Dart** and **33 Kotlin** tests pass. None of them touch this: `android/build.gradle` has `junit` only, and `VolumeService` needs a `Context`, an `AudioManager` and a live `Looper`.

**Two things are unverified:**

1. **The leak itself was never reproduced.** #444 was traced from source. I said in the issue it deserved a device check before anyone wrote the fix, and the phone came off the desk before I could do it.
2. **The fix therefore has no verification at all** beyond compiling and not regressing the existing suites.

The device repro is straightforward once a phone is attached: queue two alarms where the first sets `volumeEnforced: true` and the second sets no volume, stop the first, then move the volume rocker while the second rings. Before this change it snaps back to the first alarm's target; after it should not. Worth also re-running #437's five-alarm crash repro, since this moved the cancel that fixes it.

Draft until both are done.


gdelataillade:
Device QA done on the Pixel 8a — this is no longer unverified, and it's ready for review.

**The leak reproduced, and my first attempt at measuring it was wrong.** I tried reading the alarm stream volume with `cmd media_session volume --get` and saw it snap back to the enforced level, which looked like a clean reproduction. It wasn't: a control with nothing ringing showed that setting 7, 3 and 6 all read back as 1, so the command wasn't touching the speaker route at all and I was measuring nothing. Retracted that and observed the runnable directly instead, with a temporary tick log.

| | pre-fix (main) | post-fix |
|---|---|---|
| ticks after the queue promotion | **8** | **0** |
| successor still ringing | yes | yes |

Setup: 9201 enforcing a low volume rings, 9202 queues with no volume at all, then 9201 is stopped. Pre-fix its runnable kept forcing the stream through 9202's ring; post-fix it stops at the transition and 9202 rings unaffected.

**#437 re-verified on this branch**, since this moved the cancel that fixes it: the five-alarm crash repro gives 4 promotions, same pid, 0 NPEs.

Gate: analyze clean, 115 Dart, 33 Kotlin.

Still true, and worth keeping in mind on review: none of this is covered by an automated test. `VolumeService` needs a `Context`, an `AudioManager` and a live `Looper`, and the suite is plain JUnit.

Issue #439, opened 2026-09-05
[Android] Rings are armed with setExactAndAllowWhileIdle, which Doze rate-limits to once per nine minutes — setAlarmClock is the API exempted for alarms


samyak1409:
Every ring goes through `AlarmScheduler.setExactAlarm`, which uses `setExactAndAllowWhileIdle` (falling back to `setAndAllowWhileIdle` when the exact-alarm permission is missing, and `setExact`/`set` below API 23). `setAlarmClock` appears nowhere in the package.

Android documents those as different tiers, and the difference is specifically about alarms — *Optimize for Doze and App Standby*:

> Neither `setAndAllowWhileIdle()` nor `setExactAndAllowWhileIdle()` can fire alarms more than once per nine minutes, per app.

> Alarms set with `setAlarmClock()` continue to fire normally. The system exits Doze shortly before those alarms fire.

Two things follow. The nine-minute throttle applies to the API the package uses for **every** ring, and `setAlarmClock` is the only one the system actively wakes *ahead of* rather than merely permitting.

## Why I think this is worth revisiting rather than just noting

On #365 the standing answer is:

> Since I'm relying on the native `AlarmManager` API, this latency is outside of the plugin's control.

That's true of `setExactAndAllowWhileIdle` specifically, but the choice of which `AlarmManager` API to call is the plugin's, and it's the choice the documentation says governs exactly this behaviour. So the reasoning generalises further than it should: a latency question can have a cause inside the package.

To be clear about what I am *not* claiming: #365's own reproduction is "RING NOW" in the example app, which is a near-immediate alarm and so doesn't cleanly exercise the Doze throttle this issue is about — whatever that reporter hit, it may well be something else. This issue rests on the documented behaviour of the two APIs, not on that report. I'm citing #365 only because it is where the API question was already raised and where the general answer stands.

@anujjainwork already suggested this on #365 on 2025-08-19 — "Use AlarmManager.setAlarmClock instead of AlarmManager.setExactAndAllowWhileIdle" — as a one-liner with no argument attached, and it went unanswered. I'm reopening it as its own thread with the documentation behind it, rather than reviving a device-specific latency report, but the suggestion is theirs.

## What would change, and what to decide

[...]

Issue #365, opened 2025-05-29
Delay in delivering the notification


iDarkQ:
**Alarm plugin version**
5.1.4

**Describe the bug**
In the example app, if I click on "RING NOW" the time between alert delivers and my click varies a lot. Sometimes I don't need to wait at all, other times I have to wait 10 seconds, or even a minute.

**To Reproduce**
*Steps to reproduce the behavior:*
1. Open Example App
2. Click "RING NOW"

**Expected behavior**
It should ring either consistently (e.g. every 1 second), or at least always with low delay

**Screenshots**
https://streamable.com/2ngj5m <--- *The video will expire in two days from creation of this issue*
In this video you can see in the end I had to wait about 14 seconds for the alarm, however as I mentioned above, there are times I have to wait over a minute.

**Device info**
Xiaomi POCO X3 Pro MIUI 13.0.3 (Android 12)

**Additional context**
*Add any other context about the problem here.*

anujjainwork:
Use AlarmManager.setAlarmClock instead of AlarmManager.setExactAndAllowWhileIdle

PR #446, opened 2026-09-10
fix(android): ignore an alarm delivery that storage no longer matches (#440)

### LINKED ISSUE #440: [Android] An alarm due within 5 s is armed with Handler.postDelayed, which dies with the process and cannot be cancelled by Alarm.stop()
`AlarmScheduler.schedule` splits on a five-second threshold:

```kotlin
const val IMMEDIATE_THRESHOLD_MILLIS = 5000L
...
val armed = if (delayInMillis <= IMMEDIATE_THRESHOLD_MILLIS) {
    Handler(Looper.getMainLooper()).postDelayed(
        { context.sendBroadcast(intent) },
        delayInMillis.coerceAtLeast(0L)
    )
    true
} else {
    armAlarmManager(context, intent, alarm.dateTime.time, alarm.id)
}
```

The KDoc above it already states the cost plainly — the fallback "lives only as long as the process and cannot be cancelled through `AlarmManager.cancel`, so it is unusable for anything that has to survive termination" — and `requireDurable` exists to refuse it.

But `requireDurable` defaults to `false`, and only two internal callers pass `true`: `AlarmService.deferRefusedRing` and `SnoozeCoordinator`. The public path — `AlarmApiImpl.setAlarm`, i.e. every `Alarm.set()` — takes the default. So an app scheduling a ring three seconds out gets the weaker guarantee, is told `true`, and has no way to find out.

## The part that isn't just "less durable"

**That alarm cannot be cancelled at all.** `schedule` constructs a throwaway `Handler` and keeps no reference to the runnable, so nothing can ever call `removeCallbacks` on it. `AlarmApiImpl.stopAlarm` cancels a `PendingIntent` that was never armed:

```kotlin
AlarmService.instance?.handleStopAlarmCommand(id)
cancelPendingBroadcast(id)      // cancels an AlarmManager entry that doesn't exist
alarmIds.remove(id)
AlarmStorage(context).unsaveAlarm(id)
```

and `AlarmReceiver`'s ring path reads `AlarmSettings` from the intent extras without consulting storage, so unsaving doesn't stop it either. Net effect: `Alarm.stop(id)` completes successfully, `Alarm.getAlarms()` no longer lists the alarm, and it rings anyway a couple of seconds later. Same for `Alarm.stopAll()`.

So below the threshold, `set` returns a weaker promise than it claims and `stop` silently doesn't work.

## The actual question

Why is the fallback there at all?

`AlarmManager` accepts a trigger time three seconds out and delivers it; there's no documented floor that this is working around, and the alarm path already tolerates the millisecond-level imprecision `AlarmManager` can add. The threshold reads like it was chosen to avoid a latency worry, but it trades a small, bounded delay for an alarm that doesn't survive the process and can't be recalled — which for an alarm package seems like the wrong side of the trade.

If there is a real reason — an OEM that drops near-immediate alarms, something learned on hardware — it's worth a comment naming it, because from the outside the constant looks like a defensive guess.

## If there is no reason, a shape that keeps the intent

The version I'd suggest: **always arm `AlarmManager`, and use the in-process timer only as a backstop, never as a replacement.** Arm the durable alarm as normal; if the delivery hasn't arrived by shortly after the due time, let a `Handler` fire the same broadcast. That keeps whatever responsiveness the threshold was protecting while nothing loses durability or cancellability, and `requireDurable` stops needing to exist.

**One prerequisite, and it's a real one:** `AlarmService.onStartCommand` has no same-id guard. With `allowAlarmOverlap: false` and `allowSameSecondScheduling: true`, a second delivery of an id that is already ringing is *queued* and rings again once the first is stopped. A backstop that races a delivery which then arrives would double-ring through exactly that path, so a "have I already delivered this id" check has to land first — arguably worth having on its own merits, since a duplicate broadcast from any source has the same effect today.

There is also an interaction worth knowing about before either is decided: #441 asks for `Alarm.set` on an existing id to replace in place, where `FLAG_UPDATE_CURRENT` would allow it, rather than cancelling and recreating. An in-place replace whose new time landed in this branch would arm the `Handler` while the old `AlarmManager` entry was still live, and the alarm would fire twice — so whichever way this branch goes constrains that one.

The narrower fix, if the design change isn't wanted, is to flip `requireDurable` to default `true` so the weak path is opt-in rather than opt-out. That doesn't fix the uncancellable window for callers who opt in, but it stops `Alarm.set` handing it out silently.

## Our side

Two weather-conditional alarm apps on 5.11.0. One of them arms a "ring late" alarm relative to whenever a background wind check finishes, and we picked a 15-second lead specifically to stay out of this branch.

**That lead turned out not to be a guarantee, which is worth reporting as part of the case.** The lead is stamped from a clock frozen at the start of the evaluation pass, because the play window we judge the wind against hangs off the ring time and the fetch needs the window — so the whole HTTP fetch is spent *inside* the lead. On a cold background isolate a slow multi-model request eats fifteen seconds without difficulty, and the alarm then lands under the threshold with nothing anywhere reporting that it did. We have since had to add an explicit floor at our own platform boundary, clamping the time immediately before `Alarm.set` against the real clock.

That is the shape of the problem for a consumer: the threshold is invisible, `set` reports success, `stop` reports success, and the only way to find out is to read `AlarmScheduler.kt`. A caller-side clamp is a workaround, not a fix — it cannot help the cancellation hole for anyone who genuinely does want to ring in three seconds.

Happy to write the PR once you've said which shape you want — the backstop design with the same-id guard, or the narrower `requireDurable` default — with Kotlin tests either way.

gdelataillade:
Closes #440's cancellation hole. **Leaves the 5-second threshold itself alone** — see Scope.

## The problem

An alarm due within 5 s is armed with `Handler.postDelayed` and no reference to the runnable is kept, so nothing can cancel it. `Alarm.stop(id)` cancels a `PendingIntent` that was never armed, unsaves storage, reports success — and the alarm rings anyway, because the delivery carries its `AlarmSettings` in the intent extras and `onStartCommand` never consulted storage before ringing. Same for `stopAll()`.

The reporter hit this for real: their app stamps a 15 s lead from a clock frozen before an HTTP fetch, so a slow request drops the alarm under the threshold with nothing reporting that it did.

## The fix

`AlarmScheduler.schedule` writes the **same settings object** to storage and into the intent extras in one call, on every path including snooze and the #424 defer. So "the extras match the stored alarm at arm time" is an invariant the code already maintains, and a mismatch at delivery time means the delivery is out of date — either the alarm was stopped, or it was re-armed for a different time and this is the superseded arming.

`AlarmService.isCurrentDelivery(delivered, stored)` decides that. It is on the companion object so plain JUnit can reach it, following `BootReceiver.isStale` from #432.

Two call sites:
- **`onStartCommand`**, after the settings are parsed and before the overlap/queue branch, so a stale delivery is not even queued. It still discharges the foreground obligation — a stale delivery arrived through `startForegroundService`, and skipping that is what produced `ForegroundServiceDidNotStartInTimeException` during the #424 QA.
- **`triggerNextQueuedAlarm`**, replacing its existence-only storage check, so a queued alarm that was re-set while waiting no longer rings at its stale time.

Nothing is unsaved and nothing is reported to Dart: if the alarm was stopped there is no row to remove, and if it was superseded the new arming stands.

Deliberately **not** called `isStale` — #418 already uses that for "too late to be worth ringing", a different question with a configurable window.

## Scope

Whether the sub-5 s branch should exist at all needs either the reporter's backstop design or a `requireDurable` default flip. Both collide textually with #439 in the same function, and #439 is a product decision with a device test attached. That question stays with #365/#439; this is the part that needs no decision.

This also unblocks #441 rather than implementing it — the guard is what makes a surviving old delivery harmless, which is the hazard #441 named.

## Testing

**4 new Kotlin tests, 37 total.** The predicate is pure, which is the point of putting it on the companion object:

- a delivery matching the stored alarm rings
- a delivery for an alarm no longer stored is refused (the `Alarm.stop` case)
- a delivery superseded by a re-arm is refused
- a re-arm one millisecond later is still a different alarm

Teeth-checked both ways: making the predicate accept everything fails the three refusal tests; giving it a one-second tolerance fails only the millisecond one.

`flutter analyze` clean, `dart format` clean, 115 Dart, 37 Kotlin.

**Not covered:** the ~5 lines of wiring at each call site. `AlarmService` needs a `Context` and the suite is plain JUnit with no Robolectric — the gap already recorded for `AlarmScheduler`, `BootReceiver` and `SnoozeCoordinator`.

**No device verification yet.** The phone came off the desk before I could run it. The pass I'd want: an alarm 3 s out then stopped immediately (silent, was ringing anyway), an alarm 3 s out re-set for later (old runnable suppressed), a queued alarm re-set while waiting — and the regression row that matters most, that an ordinary alarm, a snooze and the #424 boot-window defer all still ring, since each re-arms with a new stored time and its own delivery must still match. Watch logcat for `ForegroundServiceDidNotStartInTimeException` throughout.

## Version

The CHANGELOG bullet goes under the existing unpublished heading. #445 relabels that section to 5.13.0; whichever of the two lands second reconciles the heading, so this branch does not touch `pubspec.yaml`.


gdelataillade:
Device QA done on the Pixel 8a, reproducing each case before the fix.

**The cancellation hole (case 1)** — alarm 3 s out, stopped ~1 s later:

| | pre-fix | post-fix |
|---|---|---|
| rings after being stopped | **yes**, plus the severe `settings object could not be found` | no |
| alarm players started | 1 | 0 |
| `ForegroundServiceDidNotStartInTime` | — | **0** |

That last row is the one I was most worried about: the skip path arrives via `startForegroundService`, so it has to discharge the obligation. It does.

**The re-set (case 2)** turned out worse pre-fix than I predicted, and the shape is worth recording. I expected a double ring; what actually happens is the *stale* delivery rings at the old time and the *legitimate* re-arm is then discarded as `An alarm is already ringing. Ignoring new alarm with id: 9101`. So the alarm rang at the wrong moment and the correct arming was thrown away. Post-fix: the stale delivery is suppressed and the new arming rings at its own time, with that discard message gone.

**Regressions, which matter more than the fix — all still ring, none suppressed:**

- ordinary alarm on the `AlarmManager` path
- a snooze, which re-arms to a new stored time (`snoozed until 1789052775087`, rang at exactly that)
- the **#424 boot-window defer**: rebooted with an alarm due in the window, saw the first FGS start attributed `code:BOOT_COMPLETED` and the retry 30 s later attributed `code:ALARM_MANAGER_WHILE_IDLE`, and the deferred ring landed — with **0** out-of-date suppressions

**One case I did not test**, rather than let it look covered: a queued alarm re-set *while queued*, which is the new behaviour in `triggerNextQueuedAlarm`. The predicate itself is unit-tested for exactly that (the superseded case), and the two lines of wiring mirror the other call site, but the end-to-end path is unverified.

Gate: analyze clean, 115 Dart, 37 Kotlin.

PR #447, opened 2026-09-10
fix(android): honour volumeEnforced when no volume is given (#438)

### LINKED ISSUE #438: [Android] volumeEnforced is silently unreachable with volume: null, so "leave the volume alone, but don't let it be lowered mid-ring" can't be expressed
`VolumeSettings.fixed(volume: null, volumeEnforced: true)` compiles, is a reasonable thing to write, and does nothing at all.

## Why

`AlarmService.ringAlarm` only calls `setVolume` when a volume was given:

```kotlin
if (alarmSettings.volumeSettings.volume != null) {
    volumeService?.setVolume(
        alarmSettings.volumeSettings.volume,
        alarmSettings.volumeSettings.volumeEnforced,
        ...
    )
}
```

and `startVolumeEnforcement()` is reached only from inside `setVolume`. So with a null volume the flag is read by nothing. No warning, no log line — enforcement is simply absent.

## Why that leaves a real gap rather than a cosmetic one

`volume: null` is documented as "If left unspecified or set to `null`, the current system volume at the time of the alarm will be used", and it is the right default for an alarm app with no opinion about loudness. We learned that the hard way: our dawn alarm used to pass `1.0` with enforcement on, and it turned a deliberately quiet phone up to maximum and pinned it there. Caught on a device, fixed by passing null.

But the two settings are answering different questions, and only one of them is about loudness:

- `volume` — **what** the alarm should ring at. We genuinely have no opinion; the user's slider is the answer.
- `volumeEnforced` — **whether the volume can be changed while the alarm is sounding.** For an alarm, "no" is the obvious answer regardless of who chose the number.

Today those are welded together: opting out of choosing the volume also opts you out of protecting it. So an alarm set to the user's own volume can be silenced with the volume rocker while it rings — which is exactly the failure an alarm package exists to prevent, and there is no combination of the current settings that avoids it.

## Proposed behaviour

Treat `volumeEnforced: true` with `volume: null` as **"pin the stream at whatever it reads when the ring starts."**

Mechanically that is what `setVolume` already does with a target equal to the current volume: `previousVolume` and `targetVolume` are both the current level, `setStreamVolume` is a no-op, `restorePreviousVolume` restores an unchanged value, and the enforcement loop does the only work that matters. So it is close to reordering what is there — read the stream volume as the target when none was supplied, then take the existing enforcement path.

**No special case when the stream reads 0.** If the user or the system has the alarm stream at zero, that is a user- and system-level decision, and an app has no business overriding it. Pin it at zero and let it be quiet.

## One interaction worth knowing about

This makes enforcement reachable for a set of apps that don't use it today, so it should land **after** (or together with) the orphaned-`volumeCheckRunnable` NPE in #437 — otherwise it widens the exposure to that crash.

## Our side

Weather-conditional alarms, two apps, `alarm` 5.11.0. One passes a computed volume (a wind-driven ramp) and enforces it; the other has no opinion and passes null, and is the one that can currently be turned down mid-ring. The alternative for us is a MethodChannel that reads `AudioManager.getStreamVolume(STREAM_ALARM)` and hands the number straight back to the plugin, which works but is a workaround for an expressiveness gap rather than anything an app should be carrying.

Happy to write the PR, with Kotlin tests, if you're happy with the shape.

gdelataillade:
Closes #438. Device-verified before and after on a Pixel 8a.

## The gap

`VolumeSettings.fixed(volume: null, volumeEnforced: true)` compiles, is a reasonable thing to write, and does nothing. Enforcement is reachable only from inside `setVolume`, which `ringAlarm` calls only when a volume was supplied — so with a null volume the flag is read by nothing, silently.

The two settings answer different questions: `volume` is *what* to ring at, `volumeEnforced` is *whether the level can be changed mid-ring*. Welding them together means opting out of choosing the volume also opts you out of protecting it, so an alarm at the user's own volume can be silenced with the volume rocker.

**This is an Android/iOS parity bug, not a feature request.** iOS has done the right thing since its rewrite — `targetSystemVolume = self.getSystemVolume()` when no volume is given, with `if volumeSettings.volumeEnforced` as an independent check ([`AlarmRingManager.swift:42-55`](https://github.com/gdelataillade/alarm/blob/main/ios/alarm/Sources/alarm/services/AlarmRingManager.swift#L42-L55)). Our own docs promise it too: `volume` says a null value means "the current system volume at the time of the alarm will be used".

## The change

`VolumeService.enforceCurrentVolume(...)` pins the level the stream already reads, without changing it — no `setStreamVolume`, no `savedVolume` capture, since nothing is modified and there is nothing to restore. iOS draws the same line, setting `previousVolume` only on the branch that moved the volume.

The target selection is a pure function on the companion object so the JVM suite can reach it, following `AlarmService.isCurrentDelivery` (#446) and `BootReceiver.isStale` (#432). Two rules, both non-obvious:

- **Prefer the level saved for this burst over the live stream.** At a queue promotion the stream is still raised by the alarm that just stopped, so reading it live would pin a *predecessor's* forced level. This is the same per-burst state #444 introduced.
- **Skip a muted stream.** The issue argues for pinning zero and letting it be quiet. I think that inverts the purpose: enforcement pushes the level back every second, so pinning zero leaves an alarm that is silent *and* unraisable, fighting a user trying to hear it. `volume: null` means no opinion, and that should not become "silent forever". An explicit `volume: 0.0` is untouched and still pins, because that is a deliberate choice.

The second point means **Android now deliberately diverges from iOS on muted streams**, which pin at zero. Android's rule is the better one, so iOS should follow — noted as out of scope in #438 and worth its own issue.

## Testing

**5 new Kotlin tests, 42 total**, teeth-checked both ways: ignoring the saved level fails the two tests that depend on it, and pinning a muted stream fails the two zero tests.

**Device QA**, using the tick-log harness from #444 — it reports both whether enforcement runs and what level it pins, which is what these cases turn on. Setup: 9401 rings with an explicit 0.57 and enforcement, 9402 queues behind it with no volume and enforcement on, then 9401 is stopped. The user's own level was 1; 9401 forces 4.

| | pre-fix | post-fix |
|---|---|---|
| explicit volume, enforced | ticks `target=4` | ticks `target=4` |
| null volume, enforced (promoted) | **0 ticks** | ticks **`target=1`** |

`target=1` rather than `4` is the result that matters: it is the user's own level, not the predecessor's forced one, which is exactly what the saved-level rule exists for and what reading the live stream would have got wrong.

**Not verified on device: the muted case.** This Pixel reports `Min: 1` for `STREAM_ALARM`, so the stream cannot reach 0 through the normal path and the case is unreachable here. Its only coverage is the two unit tests.

Also not covered, as usual: the ~8 lines of wiring. `VolumeService` needs a `Context`, an `AudioManager` and a live `Looper`, and the suite is plain JUnit.

Gate: analyze clean, `dart format` clean, 115 Dart, 42 Kotlin.

## Release

Rides the unpublished 5.13.0. It is a behaviour change for any app already passing `volumeEnforced: true` with a null volume — they get enforcement where they had none — hence its own CHANGELOG line.


gdelataillade:
Both real, both fixed in 3c07816 — thanks, the second one was a regression I introduced.

**P1 — restore before skipping.** You're right that the guarantee wasn't kept: at a promotion the stream still holds the predecessor's forced level, so returning without touching it meant a user whose own level was 0 got the previous alarm's volume rather than their silence. The user's level is now restored on both paths, before the decision to pin. That also removed an up-to-a-second window I hadn't noticed, where a promoted alarm rang at the predecessor's level until the first tick pulled it down.

**P2 — don't prefer the saved level under overlap.** Also right, and this one was a regression rather than a gap: pre-#438 a null-volume alarm did nothing, so it couldn't quieten anything. Preferring the saved level made an alarm with no opinion override one that explicitly asked to be loud. The saved level is now used only when nothing else is playing, which the caller can determine because this alarm's own player starts later in `ringAlarm` — so `getPlayingMediaPlayersIds()` there sees only *other* alarms: empty on a promotion, non-empty on an overlap.

**Verified on device, both directions:**

| | ticks | stream |
|---|---|---|
| overlap: explicit 0.57 (level 4) + null-volume enforced | `target=4` throughout, 2 players | 4 while ringing, restored to 1 after |
| promotion: same pair, predecessor stopped | `target=1` | already 1, no longer a delayed correction |

 under overlap and  under promotion is the distinction the gate exists for; before the fix both were 1.

43 Kotlin tests now, with the overlap case added and teeth-checked — removing the gate fails that test alone.

**Still not verified on device: P1's saved-mute case.** This Pixel reports `Min: 1` for `STREAM_ALARM`, so the stream cannot reach 0 through the normal path. Its coverage is the unit test asserting the target is 0 there; the restore-then-skip wiring around it is untested, like the rest of the wiring in this file.

